PR #1, opened 2017-01-10
Add main structure + password change feature

CPatchane:
This PR add main app functional structure and feature to change password.

* [x] Add SettingsStore
* [x] Notifier component
* [x] Password strength computing library
* [x] Add account container and stateful component for PassphraseForm component
* [x] Use mainly cozy-ui styles
* [x] Handle error for 'pasword too weak' and 'server error'
* [x] Add EN locales
* [x] Get stack domain by using a global variable (will be useful when we will be able to grab the domain from the stack)
* [x] Handle errors for passphrase too weak and current passphrase wrong


src/components/Notifier.jsx
```diff
@@ -0,0 +1,111 @@
+/** @jsx h */
```
goldoraf: This component is now used in 2 different apps, we should move it in cozy-ui.

src/components/PassphraseForm.jsx
```diff
@@ -0,0 +1,56 @@
...
+        onClick={currentPassphrase.toggleVisibility}
+        class={styles['visibility']}
+      >
+        {currentPassphrase.visible ? 'Hide' : 'Show'}
+      </a>
+    </label>
+    <input
+      type={currentPassphrase.visible ? 'text' : 'password'}
+      placeholder={t('AccountView.password.current_placeholder')}
+      value={currentPassphrase.value}
```
goldoraf: I'd prefer a lighter {...currentPassphrase} here, what do you think?

src/lib/stateFulPassphraseForm.jsx
```diff
@@ -0,0 +1,108 @@
+/*
```
goldoraf: Mybe it's time to move statefulForm into cozy-ui ;)

src/main.jsx
```diff
@@ -7,45 +7,56 @@ import { render } from 'react-dom'
...
+    <Provider store={store}>
+      <I18n context={context} lang={lang}>
+        <Router history={hashHistory}>
+          <Route component={(props) =>
+            <App {...props} />}
+          >
+            <Redirect from='/' to='account' />
+            <Route
+              path='account'
+              component={(props) =>
```
goldoraf: You can simply write `<Route path="account" component={AccountManagement} />` here. The usage of the `component={props => <AccountManagement {...props} />}` syntax should be restricted to edge cases.

goldoraf:
Great job :tada:

src/components/PassphraseForm.jsx
CPatchane: I found this way more explicit here and meaningful, even it could be lighter in your way. What do others think? @gregorylegarec ? @m4dz ? @GoOz ?

src/lib/stateFulPassphraseForm.jsx
CPatchane: This component is very specific to the passphrase form, it's why it's not the same than the statefulForm from myaccounts (which is specific to structures using a `fields` property like a connector). So, I don't think it's very useful to put those in cozy-ui as they can be very specific.

PR #7, opened 2017-02-01
add cozy-bar to app

ysa000:


src/main.jsx
```diff
@@ -2,6 +2,8 @@ import 'babel-polyfill'
 
 import './styles/main'
 
+import 'cozy-bar'
```
CPatchane: According to @m4dz, you can import cozy here:

```javascript
import cozy from 'cozy-bar'
```

vendor/assets/manifest.json
```diff
@@ -0,0 +1,18 @@
...
+        {
+            "src": "/android-chrome-512x512.png",
+            "sizes": "512x512",
+            "type": "image/png"
+        }
+    ],
+    "theme_color": "#ffffff",
+    "background_color": "#ffffff",
+    "display": "standalone"
+}
```
CPatchane: I think a new line at the end here would be better.

CPatchane:
Except these few remarks, it seems good to me, nice job! 👍

CPatchane:
Thanks @ysa000!

PR #8, opened 2017-02-01
remove extra border

ysa000:
Double border between the cozy-bar and the app's body
Removed the extra border from the app's body


GoOz:
👍

PR #10, opened 2017-02-10
[feat] Make settings great again (or work again actually)

gregorylegarec:
Cozy-settings was not working with the last version of cozy-stack which requires permission management.


goldoraf:
:100: :1st_place_medal:

PR #11, opened 2017-02-10
# Improve UI/UX

gregorylegarec:
The main goal is to avoid the feeling that the _save_ button's role is to save the whole form. To achieve that, we can:

 * Switch locale and password position, to put password at the end and to not have a button in the middle of the form.
 * Change the button label to make it more explicit
 * Avoid to center the button, we may make its width to be 100%
 * Use a secondary button instead as a primary one, as the submit action does not submit the whole form
 * Switch position between button and _Forgot your passphrase_ link, to make the button more related to the old/new passphrases fieds

capture d ecran 2017-02-10 a 19 06 20

(with @kelukelu seal of approval)

### DIFF
--- a/src/styles/fields.styl
+++ b/src/styles/fields.styl
@@ -18,6 +18,15 @@
     .password-reset-link
         margin-top      .5em
 
+    .coz-form-controls
+        padding 0
+        margin  1em 0
+        width   100%
+
+        [role=button]
+            margin-left 0
+            margin-right 0
+
     .set-field
         &:after
             opacity 0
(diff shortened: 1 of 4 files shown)


GoOz:
If master @kelukelu approved all of this, who am I to discuss the bright side of the common sense? 😛

src/styles/fields.styl
```diff
@@ -18,6 +18,15 @@
     .password-reset-link
         margin-top      .5em
 
+    .coz-form-controls
+        padding 0
+        margin  1em 0
```
GoOz: Just review the order of the properties, from the outside to the inside, in this case : 
margin -> width -> padding
and then , just align the values and we're good ^_^

GoOz:
👍

PR #12, opened 2017-02-20
[fix] Inject cozy bar with the stack

y-lohse:
Settings was still using the "old" way to include the cozy-bar and cozy-client-js, so things were a bit broken. This patch only fixes the bar, I'll submit a separate patch to use cozy-client-js if this one is ok.

@m4dz Aside from the changes included, is there anything in the webpack build process that I should change? Or something else I forgot?


m4dz:
Perfect :+1:, thanks!

PR #14, opened 2017-02-23
New view : Connected Devices

y-lohse:
This PR adds a new view, listing the connected devices.

Aside from the core view, I've copied over the Empty and Loading components from Photos, and I think they are candidates for an integration in cozy-ui. i've also switched to the Alerter system used in files. Finally, the tables style is heavily inspired by the one from Files and both should probably be merged at some point.


src/containers/Devices.jsx
```diff
@@ -0,0 +1,12 @@
+import { connect } from 'react-redux'
```
CPatchane: Good idea to split the view part and the container elements I think here, but is it really better to maintain with many containers? What do the others think? @goldoraf? @gregorylegarec?

y-lohse: To be honest, I don't think it is. As you say, I think it would be easier to maintain with the view and the container as one entity.
However, this is how the existing Accounts view is structured, and I think both should be consistent. I'm quite in favor of merging `Devices` and `DevicesView`, but then the same needs to be done for `Account` and `AccountView`. I don't mind doing them both.

src/styles/devices.styl
```diff
@@ -0,0 +1,15 @@
+@import 'cozy-ui'
+
+.devices-view
+    h2
```
GoOz: I think h2 from the account view and the device view should share the same style, and a common class, but there's some serious shared style to do here. But maybe not for now.

src/styles/devices.styl
```diff
@@ -0,0 +1,15 @@
...
+        horizontally-padded 2em
+
+    .coz-btn--revoke
+        color dodger-blue
```
GoOz: alignment and order are wrong here.
from the outside to the inside which gives paddings > bg > color

src/styles/empty.styl
```diff
@@ -0,0 +1,32 @@
+@require 'cozy-ui'
+
+.set-empty
+  text-align center
+  &:before
```
GoOz: i like a blank line before every new depth level, just to be more readable.

src/styles/empty.styl
```diff
@@ -0,0 +1,32 @@
...
+    content     ''
+    display     block
+    margin      em(100px) auto 0
+    width       em(160px)
+    height      em(140px)
+    background  embedurl("../assets/icons/icon-main-app.svg") center center / cover no-repeat
```
GoOz: Single quotes  👅

src/styles/empty.styl
```diff
@@ -0,0 +1,32 @@
...
+  h2
+    font-size(24px)
```
GoOz: wrong alignment and order here

src/styles/table.styl
```diff
@@ -0,0 +1,64 @@
...
+
+.set-content-table
+    text-align left
+    color      grey-11
+
+.set-content-header
+.set-content-cell
+    box-sizing  border-box
+    padding     em(15px) em(32px)
+    font-size(14px)
```
GoOz: wrong alignment

src/styles/table.styl
```diff
@@ -0,0 +1,64 @@
...
+    align-items     center
+    width           100%
+    border-bottom   1px solid grey-09
+
+.set-content-primary
+    color grey-08
+
+.set-content-header
+  padding         em(8px, 12px) em(28px, 12px)
+  font-size(12px)
```
GoOz: wrong alignment

src/styles/table.styl
```diff
@@ -0,0 +1,64 @@
...
+
+.set-content-primary
+    flex 0 0 40%
+
+.set-content-header.set-content-primary
+    padding-left   50px
+
+.set-content-cell.set-content-primary
+    padding-left   100px
+    font-size(16px)
```
GoOz: wrong alignment

src/styles/table.styl
```diff
@@ -0,0 +1,64 @@
...
+
+.set-content-secondary
+    flex 0 0 15%
+
+.set-loading
+    @extend $icon-spinner-small-blue
+    margin-left 0.5em
+
+.set-device-laptop
+  background embedurl("../assets/icons/icon-device-laptop.svg") 50px center no-repeat
```
GoOz: single quotes for both paths

src/styles/table.styl
```diff
@@ -0,0 +1,64 @@
...
+@media (max-width: (768/basefont)rem)
+    .set-content-header
+    .set-content-secondary
+        display none
+
+    .set-content-primary
+        flex 0 0 75%
+
+    .set-content-cell.set-content-primary
+        left-padded(61px)
```
GoOz: wrong alignment

src/containers/Devices.jsx
goldoraf: Well, the argument in favor of separate dumb component/view and its connected version is that it's easier to test.

src/styles/devices.styl
y-lohse: I agree: should be shared, not now.

y-lohse:
@GoOz thanks, all done I think. But yeah, most of these were copied over from other projects, so... 😁

PR #24, opened 2017-03-06
[chore] Upgrade to latest cozy-bar

y-lohse:
cozy-bar was declared as a dependency twice, once as a dev dependency, and once as a normal one. This PR fixes that and upgrades it to the latest available version.


GoOz:
Why are there other modules than cozy-bar added in the package.json?

y-lohse:
They are required to build the cozy-bar, unfortunately.

GoOz:
there's something i didn't get about this bar then…

y-lohse:
Well, you don't really need to declare the cozy-bar as a dependency & build it. It's only if you want to work on the bar while integrated in another app... which is what we're doing. I think.

GoOz:
Then you don't need to save the deps in the package.json, do you?

y-lohse:
Yes, you do because - again, I *think* - webpack is configured to build the bar if it's declared as a dependency. And the build fails without these extra packages. I don't even know why the builds didn't fail on travis so far.
You could argue that `cozy-bar` shouldn't be a dependency of settings at all, but we also have it as a dependency on the other apps. Then again, the extra dependencies are only dev dependencies, it's not like they make the app heavier or anything :-)

package.json
```diff
@@ -59,7 +60,9 @@
     "json-loader": "^0.5.0",
     "npm-run-all": "^3.1.0",
     "postcss": "^5.2.0",
+    "postcss-css-variables": "^0.6.0",
     "postcss-discard-duplicates": "^2.0.2",
+    "postcss-import": "^9.1.0",
```
m4dz: As @GoOz said, `cozy-bar`'s deps should not be embedded in apps but still remains in `cozy-bar` repo only. Here, the problem is that when we import `cozy-bar` in the project, webpack is looking for a `postcss.config.js` file, and it tooks the first one encoutered: the `cozy-bar`'s one.

To prevent it, we now trash the `postcss.config.js` in apps repos and put it in webpack's config file directly. See [here](https://github.com/cozy/cozy-files-v3/blob/master/config/webpack.config.base.js#L42-L46) and [there](https://github.com/cozy/cozy-files-v3/blob/master/config/webpack.config.base.js#L60-L69) (and the related [commit](https://github.com/cozy/cozy-files-v3/commit/6d25b327cbcf11eccd5a02450dc7ea75647b0d04)).

Can you do the same here? Thanks :+1: !

package.json
```diff
@@ -44,10 +44,11 @@
     "babel-runtime": "^6.0.0",
     "chai": "^3.5.0",
     "copy-webpack-plugin": "^4.0.1",
-    "cozy-bar": "beta",
+    "cozy-bar": "3.0.0-beta9",
     "cozy-client-js": "^0.1.0",
     "css-loader": "^0.26.0",
     "css-mqpacker": "^4.0.1",
+    "csswring": "^5.1.1",
```
m4dz: This one is ok and previously missing, see previous comment about embed postcss config :wink:

package.json
y-lohse: All done! Thanks for the tip.

PR #26, opened 2017-03-09
[fix] Use the domain provided by the stack

y-lohse:


gregorylegarec:
Thanks @y-lohse, you should have pushed directly into master branch 😈

PR #31, opened 2017-03-22
feat: revoke devices

lespacedunmatin:
Miss a part of CSS on the popin.


src/components/DevicesView.jsx
```diff
@@ -8,18 +8,28 @@ import { translate } from 'cozy-ui/react/helpers/i18n'
...
 const getDeviceKindClass = kind => kind === 'mobile' ? styles['set-device-phone'] : styles['set-device-laptop']
 
-const DevicesView = ({ t, f, isFetching, devices }) => (
+const DevicesView = ({ t, f, isFetching, devices, openDeviceRevokeModale, deviceToRevoke, onDeviceModaleRevoke, onDeviceModaleRevokeClose, devicePerformRevoke }) => (
   <div className={devicesStyles['devices-view']}>
     <h2>{t('DevicesView.title')}</h2>
     { isFetching && <Loading /> }
     { !isFetching && devices.length === 0 && <Empty emptyType='devices' />}
-    { devices.length > 0 && (
+    { !isFetching && devices.length > 0 && (
```
y-lohse: Why is there a new `!isFetching` condition here?

src/components/DevicesModaleRevokeView.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import { translate } from 'cozy-ui/react/helpers/i18n'
+import Modal from 'cozy-ui/react/Modal'
+
+const devicesModaleRevokeView = ({ t, device, revokeDevice, cancelAction }) => {
+  return (<Modal
+    title={t('revokeDevice.title')}
+    description={t('revokeDevice.description', {name: device.client_name})}
+    subText={t('revokeDevice.subText')}
```
y-lohse: The Modal component doesn't support a `subText` property, so it's not showing up as far as I can tell.
The correct approach here is to create the DOM structure you want as a description before hand, and pass it all in the `description` prop. [Here's an example](https://github.com/cozy/cozy-files-v3/blob/master/src/containers/DeleteConfirmation.jsx#L13) of how to do that.

src/components/DevicesView.jsx
```diff
@@ -33,7 +43,16 @@ const DevicesView = ({ t, f, isFetching, devices }) => (
             </div>
             <div className={classNames(styles['set-content-cell'], styles['set-content-secondary'])} />
             <div className={classNames(styles['set-content-cell'], styles['set-content-secondary'])} />
-            <div className={classNames(styles['set-content-cell'])} />
+            <div className={classNames(styles['set-content-cell'])}>
+              <button
```
y-lohse: The revoke button's style doesn't match the mockup -- it should be blue and aligned with the column title.
The style should still be there, the button was temporarily removed in [this commit](https://github.com/cozy/cozy-settings/commit/ee4d09cf7f9f3043a0ae1b17cf79ca4c8cdcc1e2).

src/components/DevicesView.jsx
```diff
@@ -8,18 +8,28 @@ import { translate } from 'cozy-ui/react/helpers/i18n'
...
     { isFetching && <Loading /> }
     { !isFetching && devices.length === 0 && <Empty emptyType='devices' />}
-    { devices.length > 0 && (
+    { !isFetching && devices.length > 0 && (
       <div className={classNames(
         styles['set-content-table']
       )}>
+        {openDeviceRevokeModale &&
+          <DevicesModaleRevokeView
+            cancelAction={onDeviceModaleRevokeClose}
```
y-lohse: I am unable to cancel the revokation modal. Here's the error that I get:

```
Given action "DEVICES_MODALE_REVOKE_CLOSE", reducer "deviceToRevoke" returned undefined. To ignore an action, you must explicitly return the previous state.
    at combination (combineReducers.js:135)
    at dispatch (createStore.js:179)
    at index.js:89
    at index.js:14
    at Object.onDeviceModaleRevokeClose [as click] (Devices.jsx:19)
    at HTMLButtonElement.eventProxy (preact.js:132)
```

src/components/DevicesView.jsx
lespacedunmatin: Because I run loading (set `isFetching`to `true`) while the DELETE request is sent on the network. Without that condition, the list of devices is pushed to the bottom of the page for a few seconds, it’s not nice.

src/components/DevicesView.jsx
lespacedunmatin: Yep, that’s the missing CSS part.

src/components/DevicesView.jsx
lespacedunmatin: Woops, 990ccf7 corrects that.

src/components/DevicesView.jsx
y-lohse: Ah, I see!
I think that `isFetching` is used to indicate when the initial data is being fetched, and not so much for subsequent AJAX calls. The [info and password updating](https://github.com/cozy/cozy-settings/blob/master/src/actions/index.js#L37) for example don't change that flag.
Is it really useful to change that `fetching` flag in this case? If not I would suggest to remove it.

src/components/DevicesView.jsx
lespacedunmatin: Corrected by 3661372.

src/components/DevicesView.jsx
lespacedunmatin: It brings to the user the information that the request is being processed. We could have a spinner in the modal instead.

y-lohse:
Great, thank you!

lespacedunmatin:
Woot :)

PR #36, opened 2017-03-29
Application install/update screen implementation (SET-11)

doubleface:
Here is a first implementation of the hidden application install/update screen.

This hidden screen is accessible with the following url : http://settings.cozy.local:8080/#/install

It is important that the settings application has the "settings" slug or else the stack will refuse any application install/update.

And it is also important that the cozy-stack is updated (at the moment, the docker image does not work)

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -23,6 +23,9 @@ export const DEVICE_REVOKE = 'DEVICE_REVOKE'
 export const DEVICE_REVOKE_SUCCESS = 'DEVICE_REVOKE_SUCCESS'
 export const DEVICE_REVOKE_FAILURE = 'DEVICE_REVOKE_FAILURE'
 export const ALERT_CLOSED = 'ALERT_CLOSED'
+export const INSTALL_APP = 'INSTALL_APP'
+export const INSTALL_APP_SUCCESS = 'INSTALL_APP_SUCCESS'
+export const INSTALL_APP_FAILURE = 'INSTALL_APP_FAILURE'
 
 export const fetchInfos = () => {
   return (dispatch, getState) => {
@@ -155,6 +158,34 @@ export const deviceModaleRevokeClose = () => ({
   type: DEVICES_MODALE_REVOKE_CLOSE
 })
 
+export const installApp = (slug, repourl, isupdate) => {
+  return (dispatch, getState) => {
+    dispatch({ type: INSTALL_APP })
+    let verb = isupdate ? 'PUT' : 'POST'
+    return cozyFetch(verb, `/apps/${slug}?Source=${encodeURIComponent(repourl)}`)
+    .then(response => {
+      dispatch({
+        type: INSTALL_APP_SUCCESS,
+        alert: {
+          message: `InstallView.${isupdate ? 'update' : 'install'}_success`,
+          messageData: {slug},
+          level: 'success'
+        }
+      })
+    })
+    .catch(() => {
+      dispatch({
+        type: INSTALL_APP_FAILURE,
+        alert: {
+          message: `InstallView.${isupdate ? 'update' : 'install'}_error`,
+          messageData: {slug},
+          level: 'error'
+        }
+      })
+    })
+  }
+}
+
 const STACK_DOMAIN = '//' + document.querySelector('[role=application]').dataset.cozyDomain
 const STACK_TOKEN = document.querySelector('[role=application]').dataset.cozyToken
 
@@ -181,7 +212,7 @@ const cozyFetch = (method, path, body) => {
         data = response.text()
       }
 
-      return (response.status === 200 || response.status === 204)
+      return (response.status >= 200 && response.status < 300)
         ? data
         : data.then(Promise.reject.bind(Promise))
     })
--- a/src/components/InstallerView.jsx
+++ b/src/components/InstallerView.jsx
@@ -1,0 +1,31 @@
+import viewStyles from '../styles/view'
+import React, { Component } from 'react'
+import { translate } from 'cozy-ui/react/helpers/i18n'
+import classNames from 'classnames'
+
+class InstallerView extends Component {
+  doInstall () {
+    // get the values from the fields
+    const slug = document.querySelector('#slug').value
+    const repourl = document.querySelector('#repourl').value
+    const isupdate = document.querySelector('#isupdate').checked
+    this.onSubmit(slug, repourl, isupdate)
+  }
+
+  render ({t, onSubmit}) {
+    this.onSubmit = onSubmit
+    return (
+      <div role='contentinfo'>
+        <div className={classNames(viewStyles['set-view-content'], viewStyles['set-view-content--narrow'])}>
+          <h2 className={viewStyles['set-view-title']}>{t('InstallView.title')}</h2>
+          <input id='slug' placeholder='slug' /><br />
+          <label>{t('InstallView.update')}</label> <input id='isupdate' type='checkbox' />
+          <input id='repourl' value='git://github.com/username/repository.git#branch' />
+          <button onclick={this.doInstall.bind(this)}>{t('InstallView.submit')}</button>
+        </div>
+      </div>
+    )
+  }
+}
+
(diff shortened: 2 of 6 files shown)


src/components/InstallerView.jsx
```diff
@@ -0,0 +1,31 @@
+import viewStyles from '../styles/view'
+import React, { Component } from 'react'
+import { translate } from 'cozy-ui/react/helpers/i18n'
+import classNames from 'classnames'
+
+class InstallerView extends Component {
+  doInstall () {
+    // get the values from the fields
+    const slug = document.querySelector('#slug').value
```
gregorylegarec: We try to avoid querying the DOM in component. You may instead use a stateful Component :
```javascript
class InstaerView extends Component {
  constructor (props) {
    super(props)
    this.state = {
      slug: null,
      url: null
   }

   render (props) {
     return(
      ...
      <input [...] name="slug" onChange={(event) => this.setState({slug: event.target.value})} />
      ...
  }
```
Then, in `doInstall`, we just have to rely on component's state.

@goldoraf May you confirm please ?

src/components/InstallerView.jsx
```diff
@@ -0,0 +1,31 @@
...
+  render ({t, onSubmit}) {
+    this.onSubmit = onSubmit
+    return (
+      <div role='contentinfo'>
+        <div className={classNames(viewStyles['set-view-content'], viewStyles['set-view-content--narrow'])}>
+          <h2 className={viewStyles['set-view-title']}>{t('InstallView.title')}</h2>
+          <input id='slug' placeholder='slug' /><br />
+          <label>{t('InstallView.update')}</label> <input id='isupdate' type='checkbox' />
+          <input id='repourl' value='git://github.com/username/repository.git#branch' />
+          <button onclick={this.doInstall.bind(this)}>{t('InstallView.submit')}</button>
```
gregorylegarec: We can avoid binding `doInstall` here by doing something like :
```jsx
<button onclick={(event) => { this.doInstall() }}>{t('InstallView.submit')}</button>
```

src/containers/Installer.jsx
```diff
@@ -0,0 +1,20 @@
+import { connect } from 'react-redux'
+
+import InstallerView from '../components/InstallerView'
+import { installApp } from '../actions'
+
+const mapStateToProps = (state, ownProps) => ({
+  slug: state.slug,
```
gregorylegarec: As `slug` and `url` are never set in a reducer, we can avoid relying on global state here. Those are more two properties of InstallerView's internal state. I know it's a little bit different than the way `ProfileView` works but I think that @goldoraf will confirm.

src/actions/index.js
```diff
@@ -181,7 +212,7 @@ const cozyFetch = (method, path, body) => {
         data = response.text()
       }
 
-      return (response.status === 200 || response.status === 204)
+      return (response.status >= 200 && response.status < 300)
```
gregorylegarec: What are the expected response status ? Isn't [200 : 299] too wide ?

src/components/InstallerView.jsx
```diff
@@ -0,0 +1,31 @@
...
+    this.onSubmit(slug, repourl, isupdate)
+  }
+
+  render ({t, onSubmit}) {
+    this.onSubmit = onSubmit
+    return (
+      <div role='contentinfo'>
+        <div className={classNames(viewStyles['set-view-content'], viewStyles['set-view-content--narrow'])}>
+          <h2 className={viewStyles['set-view-title']}>{t('InstallView.title')}</h2>
+          <input id='slug' placeholder='slug' /><br />
```
gregorylegarec: As we already speak we should use the `Input` component here, it should not be very expensive.

src/components/InstallerView.jsx
enguerran: See https://facebook.github.io/react/docs/forms.html#controlled-components

As React/preact uses a specific paradigm to manage states in component, we recommend to use the same paradigm for the special html elements that manage state.

src/containers/Installer.jsx
enguerran: — source [You might not need redux](https://medium.com/@dan_abramov/you-might-not-need-redux-be46360cf367)

src/actions/index.js
```diff
@@ -155,6 +158,34 @@ export const deviceModaleRevokeClose = () => ({
   type: DEVICES_MODALE_REVOKE_CLOSE
 })
 
+export const installApp = (slug, repourl, isupdate) => {
+  return (dispatch, getState) => {
+    dispatch({ type: INSTALL_APP })
+    let verb = isupdate ? 'PUT' : 'POST'
```
enguerran: Why do you declare `verb` as a mutable variable?

src/components/InstallerView.jsx
goldoraf: Yes, I confirm

src/actions/index.js
```diff
@@ -155,6 +158,34 @@ export const deviceModaleRevokeClose = () => ({
   type: DEVICES_MODALE_REVOKE_CLOSE
 })
 
+export const installApp = (slug, repourl, isupdate) => {
```
enguerran: Just a matter of syntax's homogeneity: we use camelCase for variable name: `repoURL` and `isUpdate`.

src/actions/index.js
doubleface: I just had to add 202 for app install/update result. I will just add it explicitely.

src/containers/Installer.jsx
goldoraf: Yep, I agree, local state is fine here.

src/components/InstallerView.jsx
doubleface: Thanks, still some backbone stuff remains in my head

src/components/InstallerView.jsx
```diff
@@ -0,0 +1,31 @@
...
+class InstallerView extends Component {
+  doInstall () {
+    // get the values from the fields
+    const slug = document.querySelector('#slug').value
+    const repourl = document.querySelector('#repourl').value
+    const isupdate = document.querySelector('#isupdate').checked
+    this.onSubmit(slug, repourl, isupdate)
+  }
+
+  render ({t, onSubmit}) {
```
enguerran: Passing `props` in the `render`'s parameter is not react compatible.
By the way, a props is accessible in the whole component instance.
So there is no need to `this.onSubmit = onSubmit`.

src/containers/Installer.jsx
```diff
@@ -0,0 +1,20 @@
...
+  appUrl: state.appUrl
+})
+
+const mapDispatchToProps = (dispatch, ownProps) => ({
+  onSubmit: (slug, appUrl, isUpdate) => {
+    return dispatch(installApp(slug, appUrl, isUpdate))
```
enguerran: In ES2017 arrow function, we can avoid `return` statement if we have only one instruction.

> Note that an arrow function with a block body does not automatically return a value. Use a return statement for that.
— https://hacks.mozilla.org/2015/06/es6-in-depth-arrow-functions/

```js
onSubmit: (slug, appURL, isUpdate) => dispatch(installApp(slug, appURL, isUpdate))
```

src/components/InstallerView.jsx
doubleface: ok

src/containers/Installer.jsx
doubleface: :+1:

src/actions/index.js
doubleface: ok

src/actions/index.js
doubleface: No good reason, but I am sure you knew it :-)

src/components/InstallerView.jsx
doubleface: ok

src/containers/Installer.jsx
doubleface: ok

doubleface:
Closing this because I had to rebase my branch

PR #37, opened 2017-03-29
[refacto] Dumber inputs 📝

gregorylegarec:
Quick refacto to make incoming form based features easier.

### DIFF
--- a/src/components/Field.jsx
+++ b/src/components/Field.jsx
@@ -5,14 +5,14 @@ import React from 'react'
 import classNames from 'classnames'
 import { translate } from 'cozy-ui/react/helpers/i18n'
 
-const Field = ({ t, name, submitting, saved, errors, children }) => (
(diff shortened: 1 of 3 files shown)


src/components/Field.jsx
```diff
@@ -5,14 +5,14 @@ import React from 'react'
 import classNames from 'classnames'
 import { translate } from 'cozy-ui/react/helpers/i18n'
 
-const Field = ({ t, name, submitting, saved, errors, children }) => (
+const Field = ({ t, name, label, description, submitting, saved, errors, children }) => (
```
kosssi: name is never used?

gregorylegarec: @kosssi Well spotted. I pushed a quick-clean directly on `master`.

y-lohse: I do wonder why the linter didn't catch this though.

PR #38, opened 2017-03-30
 Application install/update screen implementation (SET-11) 

doubleface:
Here is a second (see #36 ) implementation of the hidden application install/update screen.

This hidden screen is accessible with the following url : http://settings.cozy.local:8080/#/install

It is important that the settings application has the "settings" slug or else the stack will refuse any application install/update.

And it is also important that the cozy-stack is updated (at the moment, the docker image does not work)


gregorylegarec:
Thanks @doubleface and congrats for your first (in many) contributions on apps 🎉

PR #40, opened 2017-04-04
Set 6 - new password

lespacedunmatin:
Be able to request a new password on the profile page.


src/components/PassphraseForm.jsx
```diff
@@ -79,9 +89,38 @@ class PassphraseForm extends Component {
...
+        {!passphraseNewRequesting && <a
+          href='#'
+          className={styles['password-reset-link']}
+          onclick={() => this.handleForgot()}>
           {t('ProfileView.password.reset_link')}
-        </a>
+        </a>}
+
+        {passphraseNewRequesting &&
+        <span className={styles['passphrase_new_send']}>
```
GoOz: You should call for Loading component instead.
Like this https://github.com/cozy/cozy-settings/blob/master/src/components/DevicesView.jsx#L21
And then, if the default style for the Loading component isn't quite what you want, add an overwriting declaration for `.set-loading` in the `.styl` of the view.

src/components/PassphraseForm.jsx
```diff
@@ -79,9 +89,38 @@ class PassphraseForm extends Component {
...
+          {t('ProfileView.password.reset_sending')}
```
GoOz: I guess this text won't be needed if you use Loading component.

src/styles/fields.styl
```diff
@@ -76,3 +76,13 @@
                 background    emerald
             &::-moz-progress-bar
                 background    emerald
+
+    .passphrase_new_send
```
GoOz: See my previous comment on using Loading component to adjust this.

src/styles/fields.styl
```diff
@@ -76,3 +76,13 @@
...
+        margin-top .5em
+
+        &:before
+            @extend $icon-spinner-blue
+            @extend $icon-16
+            opacity 1
+            left -8px
+            position relative
```
GoOz: I forgot this. There should be a blank line at the end.

lespacedunmatin: Corrected by fbe6d84.

GoOz:
Yeah ok, after discussing the Loading issue, I see got it all wrong, just ignore my comments regarding this.

gregorylegarec:
Thanks @lespacedunmatin !

PR #45, opened 2017-05-04
Fix browser autocomplete on passwords

m4dz:


gregorylegarec:
Like @GoOz, I think that I am missing a point. The specs tells that the `autocomplete` attributes should only have values between `on|off|default` https://www.w3.org/TR/2011/WD-html5-20110525/common-input-element-attributes.html#the-autocomplete-attribute

GoOz:
@gregorylegarec I know there's some tricks to really make it work ( https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion ) I just don't get the workflow here 😛

gregorylegarec:
@GoOz Thanks, I got it now !

gregorylegarec:
The goal of this PR is to avoid password recording by the browser when the input is only a text input.

src/components/Input.jsx
```diff
@@ -47,6 +47,7 @@ export const PasswordInput = translate()(
             value={value}
             onInput={onInput}
             className={inError ? styles['error'] : ''}
+            autocomplete={autocomplete || 'off'}
```
gregorylegarec: You forgot to declare `autocomplete` line 35

GoOz:
Ooooh that was the input text part that i didn't get. ok thanks

m4dz:
In fact, `autocomplete` standard from WHAT-WG allows many more options: https://html.spec.whatwg.org/multipage/forms.html#autofill. I so picked the twos `new-password` and `current-password` as they explicitely rely on what we want, apply on _password group` (which contains both `type=text` and `type=password` and, if browsers vendors do their shit, disable the autocomplete as the _password group_ implies security issues :smiley:

m4dz:
But I can be really pervert and say that I explicitely picked a random string that reflects a semantic content :smiling_imp:

gregorylegarec:
Thanks @m4dz !

PR #46, opened 2017-05-10
SET-16 removed activated services tab

doubleface:


GoOz:
Is this view removed FOR EVER AND EVER? 😮 

Not totally sure about `.set-table-status`. Just check what the table in the device view needs and remove the unused.

doubleface:
@GoOz This view is removed from the code but this commit will get a tag so we can get back to it when needed.

And OK I will check what is needed with the device view

gregorylegarec:
@GoOz already checked, everything will be ok. I think.

gregorylegarec:
Thanks @doubleface ! 👍

PR #47, opened 2017-05-16
chore: add piwik tracker + devtools

enguerran:
Same as https://github.com/cozy/cozy-ui/pull/104

\+ devtools in development

### DIFF
--- a/src/main.jsx
+++ b/src/main.jsx
@@ -1,4 +1,6 @@
-/* global cozy */
+/* global __DEVELOPMENT__ __PIWIK_TRACKER_URL__ __PIWIK_SITEID__ */
+/* global cozy Piwik */
+
 import 'babel-polyfill'
 
 import './styles/main'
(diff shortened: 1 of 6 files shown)


enguerran:
[(2) [GLOB-7] Traces d'usage](https://trello.com/c/zUe4D0KA/269-2-glob-7-traces-d-usage)

src/main.jsx
```diff
@@ -93,10 +106,22 @@ document.addEventListener('DOMContentLoaded', () => {
...
+  try {
+    var PiwikReactRouter = require('piwik-react-router')
+    const piwikTracker = (Piwik.getTracker(), PiwikReactRouter({
+      url: __PIWIK_TRACKER_URL__,
+      siteId: __PIWIK_SITEID__,
+      injectScript: false
+    }))
+    piwikTracker.push(['enableHeartBeatTimer'])
+    history = piwikTracker.connectToHistory(hashHistory)
+  } catch (err) {}
```
gregorylegarec: Silent error ? Are we sure it's a good idea here ? At least a `console.warn()` could save a little developer mental health.

enguerran: As it is for tracking, I silent this intentionally because I don't know what we could do if an error occurs.
If you want to debug this thing, I guess you use a debugger, not a `console.log` debugging workflow. But in production, no need to display something in case of activated DNT option, or other possible cases.

PR #52, opened 2017-06-12
Add `use_tracker` preference

CPatchane:
* Fix config to get better css class debugging
* Add a ReactMarkdown wrapper component for better reusability
* Upgrade `preact`, `preact-compat` and `cozy-ui` dependencies
* Fix event handler on checkbox type input (to use onChange)
* Fix a UI bug when catching a generic server error
* Add `use_tracker` preference setting (ping @aenario or anyone else from the back), __this supposes that the stack will use the `use_tracker` property from the settings document. Let me know if that part should be changed or renamed.__
* Add CGU link in profile view


config/webpack.config.cozy-ui.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
       {
         test: /\.styl$/,
         loader: extractor.extract('style', [
-          'css?importLoaders=1&modules',
+          'css-loader?importLoaders=1&modules&localIdentName=[local]--[hash:base64:5]',
```
GoOz: 👍

package.json
```diff
@@ -81,12 +81,12 @@
   },
   "dependencies": {
     "classnames": "^2.2.0",
-    "cozy-ui": "^3.0.0-beta24",
+    "cozy-ui": "beta32",
```
GoOz: just `beta32`? I'm skeptical about this.

package.json
```diff
@@ -81,12 +81,12 @@
...
     "date-fns": "^1.27.2",
     "node-polyglot": "^2.2.0",
     "piwik-react-router": "^0.8.2",
-    "preact": "^7.1.0",
-    "preact-compat": "^3.9.0",
+    "preact": "^8.1.0",
```
GoOz: I know @goldoraf had some troubles with this version but I'm not quite sure you're concerned. Pinging him just in case.

package.json
y-lohse: Didn't we find out this format causes issues with yarn's cache or something like that?

aenario:
:+1:  Ok for use_tracker, we might need to handle it differently back-side, but the `PUT /instance` api wil stay the same !

package.json
CPatchane: Oops, you're right, I change that 👍

src/styles/fields.styl
```diff
@@ -28,20 +28,23 @@
             margin-right 0
 
     .set-field
-        &:after
+        &:not(.set-field--checkbox):after,
+        &.set-field--checkbox .set-toggle label:after
```
GoOz: I am not convinced by all of this. 
If you have to cancel then reattribute a style so many time in the same file, maybe you should not put it in the first place. 
What i'm trying to say is : don't put the `set-field` class on this component. Duplicate this class with a new one and modify instead. 
Or you can also create some kind of HOC (High Order Class 🤣) `set-field` as a wrapper and then `set-field--input` for inputs specificity and `set-field--toggle` for toggles specificity.
You know what I'm saying?

package.json
CPatchane: I juste fixed it 👍

package.json
goldoraf: Preact 8.X introduced a different way of handling scroll events, and that made react-virtualized fail at handling long lists in Drive. That's the reason I reverted Drive's preact version to 7.X

CPatchane: Ok, good to know so. But here I think that will be fine, thanks @goldoraf  and @GoOz 👍

src/styles/fields.styl
CPatchane: It's done 👍

GoOz:
Way better ❤️  
Thank you 👍

PR #53, opened 2017-06-12
feat: add a link to devices clients if no device is connected to the account

lespacedunmatin:


src/components/Empty.jsx
```diff
@@ -8,8 +9,9 @@ export const Empty = ({ t, emptyType }) => {
     <div>
       {emptyType === 'devices' &&
         <div className={styles['set-empty']}>
-          <h2>{ t('Empty.devices_title') }</h2>
-          <p>{ t('Empty.devices_text')}</p>
+          <h2>{t('Empty.devices.title')}</h2>
+          <p>{t('Empty.devices.text')}</p>
+          <p><a className={classNames('coz-btn', 'coz-btn--regular')} href={t('Empty.devices.link.href')}>{t('Empty.devices.link.text')}</a></p>
```
CPatchane: Like other external links on Cozy, it's better to open it in a new tab using `target='_blank'` to avoid leaving the platform.

GoOz: I don't really like `target="_blank"` either but that will give you feedbacks if you don't do it, indeed.

lespacedunmatin: @GoOz, there is not yet an accessibility helper for the "opens in a new window" title ?

GoOz: Not yet, no. So far we didn't had target blank anywhere and then, bam, one time, there's a need for this, quickly with no extra time. :/

CPatchane:
Thanks @lespacedunmatin 👍

PR #56, opened 2017-06-26
Handle tracking preference as string in database

CPatchane:


CPatchane:
[ETU l'option de tracking est prise en compte dans settings](https://trello.com/c/ATeWaBs9/488-etu-loption-de-tracking-est-prise-en-compte-dans-settings)

CPatchane:
@gregorylegarec We could use a boolean in the database, it's just an exception for the tracking since it's serve by the stack on the [`.ejs` file](https://github.com/cozy/cozy-settings/blob/master/src/index.ejs#L28) for the tracking enabling.

src/actions/index.js
```diff
@@ -36,6 +36,10 @@ export const fetchInfos = () => {
     dispatch({ type: FETCH_INFOS })
     cozyFetch('GET', '/settings/instance')
       .then(instance => {
+        // tracking preference is stored as string, convert it to boolean for the checkbox
+        if (instance && instance.data && instance.data.attributes && instance.data.attributes.hasOwnProperty('tracking')) {
```
gregorylegarec: Why `hasOwnProperty` ?

gregorylegarec:
@CPatchane Ok, so it looks like we are hacking a stack issue here. :D
One more question and we're done. :)

src/actions/index.js
CPatchane: Don't you think we should handle falsy value here? Like empty string?

gregorylegarec: As we only interested in a string having a `true` value, I don't think so.

gregorylegarec:
Thanks @CPatchane !

gregorylegarec:
[(?) Injecter la variable de tracking aux applications](https://trello.com/c/aHcFqunq/529-injecter-la-variable-de-tracking-aux-applications)

PR #64, opened 2017-08-07
[WIP] Add Claudy support form

CPatchane:
WIP:
⚠️ ⚠️ ⚠️ 
Need to merge these PRs before merging this one :
~~https://github.com/cozy/cozy-ui/pull/157~~
~~https://github.com/cozy/cozy-bar/pull/95~~
~~https://github.com/cozy/cozy-client-js/pull/194~~
⚠️ ⚠️ ⚠️ 
When released, the manifest need also to be udpated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
   },
   "dependencies": {
     "classnames": "^2.2.0",
-    "cozy-ui": "3.0.0-beta39",
+    "cozy-ui": "^3.0.0-beta40",
     "date-fns": "^1.27.2",
     "node-polyglot": "^2.2.0",
     "piwik-react-router": "^0.8.2",
--- a/src/services/Claudy.jsx
+++ b/src/services/Claudy.jsx
@@ -140,7 +140,9 @@ export class Claudy extends Component {
           <button className='coz-claudy-menu-header-back-button' onClick={this.goBack} />
         </header>
         <div className='coz-claudy-menu-content-wrapper'>
-          <div className='coz-claudy-menu-content' >
+          <div className='coz-claudy-menu-content'
+            ref={(contener) => { this.claudyContener = contener }}
+          >
             <div className='coz-claudy-menu-actions-list'>
               {claudyActions.map(action => (
                 <a className='coz-claudy-menu-action' data-complete={action.complete} onClick={() => this.selectAction(action)}>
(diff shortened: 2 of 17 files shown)


CPatchane:
[(5) [NUA-7] ETU je peux contacter le support directement depuis le nuage](https://trello.com/c/aGWhw9XW/541-5-nua-7-etu-je-peux-contacter-le-support-directement-depuis-le-nuage)

src/services/Claudy.jsx
```diff
@@ -140,7 +140,9 @@ export class Claudy extends Component {
           <button className='coz-claudy-menu-header-back-button' onClick={this.goBack} />
         </header>
         <div className='coz-claudy-menu-content-wrapper'>
-          <div className='coz-claudy-menu-content' >
+          <div className='coz-claudy-menu-content'
+            ref={(contener) => { this.claudyContener = contener }}
```
gregorylegarec: I think there is a typo here ? s/contener/container/

src/services/ClaudyActionComponents/Support.jsx
```diff
@@ -14,13 +21,21 @@ export class Support extends Component {
...
+      })
+    }
+    nextProps.opened !== this.props.opened && nextProps.opened
       ? this.onOpen() : this.onReturn()
   }
 
   resize () {
     typeof this.props.resizeIntent === 'function' && this.props.resizeIntent({
-      height: 27.5 * 16 // 27.5em
+      height: (31.5 * 16) // 31.5em
```
gregorylegarec: I think I should have precised height in `px` directly IMHO, and kept the comment.

CPatchane: I think that doesn't cost anything here and it will easily force us to resize only using `em` here. We won't be attempted to use pixel directly. What do you think?

src/services/Claudy.jsx
CPatchane: Oops, you're completely right! 👍

gregorylegarec:
Thanks @CPatchane !

package.json
```diff
@@ -83,7 +83,7 @@
   },
   "dependencies": {
     "classnames": "^2.2.0",
-    "cozy-ui": "3.0.0-beta39",
+    "cozy-ui": "^3.0.0-beta40",
```
GoOz: You shouldn't write the `^` before the version, it causes an issue when you update with yarn. Well, until the beta is here at least. Yarn can't handle version with text like beta in it. For it, it's still 3.0.0 regardless the beta number.

PR #65, opened 2017-08-08
Claudy improvements

CPatchane:


CPatchane:
[(5) [NUA-7] ETU je peux contacter le support directement depuis le nuage](https://trello.com/c/aGWhw9XW/541-5-nua-7-etu-je-peux-contacter-le-support-directement-depuis-le-nuage)

src/services/ClaudyActionComponents/Support.jsx
```diff
@@ -77,12 +82,14 @@ export class Support extends Component {
                 onChange={(e) => { this.setState({message: e.target.value}) }}
               />
             </label>
-            {!isSent && !isSending && !error &&
+            {(
+              (!isSent && !isSending && !error) ||
+              (isSent && !isSending && !error && message)) &&
```
gregorylegarec: Shouldn't we be using some kind of selectors here ? The code readibility is not optimal imho.

gregorylegarec:
Thanks @CPatchane 🍨

src/services/ClaudyActionComponents/Support.jsx
CPatchane: Agree, I'll think about it if I'll need to update that so. 👍

PR #68, opened 2017-09-07
Add accounts count for cozy-collect claudy action :sparkles:

CPatchane:
@GoOz I just discover that `text-decoration` cannot be overwritten by the child elements, so I had to use some tricky code to handle that correctly (no text-decoration in my child element `<span>`) and overwrite styles from cozy-ui.

### DIFF
--- a/src/services/Claudy.jsx
+++ b/src/services/Claudy.jsx
@@ -173,11 +173,18 @@ export class Claudy extends Component {
                   <p className='coz-claudy-menu-action-title'>
                     {t(`claudy.actions.${action.slug}.title`)}
                   </p>
-                  {action.complete &&
+                  {action.complete && action.slug !== 'cozy-collect' &&
                     <img
                       className='coz-claudy-menu-action-check'
                       src={this.checkIcon}
                     />
+                  }
+                  {action.complete && action.slug === 'cozy-collect' &&
+                    <div className='coz-claudy-menu-action-count-wrapper'>
+                      <span className='coz-claudy-menu-action-count'>
+                        {claudyInfos.collectAccounts.length}
+                      </span>
+                    </div>
                   }
                 </a>
               ))}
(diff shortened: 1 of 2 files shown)


CPatchane:
[(2) [NUA-9] Avoir un compteur pour le cas d'usage des connecteurs](https://trello.com/c/EKoTuFM1/570-2-nua-9-avoir-un-compteur-pour-le-cas-dusage-des-connecteurs)

CPatchane:
@GoOz I wanted to add a `<span>`that has no `text-decoration` on it (which is a property of its parent element). But when a `text-decoration` is defined by a parent, it cannot be overwritten by the child elements.
> Source from [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration) : Text decorations are drawn across descendant text elements. This means that if an element specifies a text decoration, then a child element can't remove the decoration.

src/services/Claudy.jsx
```diff
@@ -173,12 +173,19 @@ export class Claudy extends Component {
                   <p className='coz-claudy-menu-action-title'>
                     {t(`claudy.actions.${action.slug}.title`)}
                   </p>
-                  {action.complete &&
+                  {action.complete && action.slug !== 'cozy-collect' &&
```
gregorylegarec: Shouldn't we use a constant here ?

src/services/Claudy.jsx
```diff
@@ -173,12 +173,19 @@ export class Claudy extends Component {
...
                     <img
                       className='coz-claudy-menu-action-check'
                       src={this.checkIcon}
                     />
                   }
+                  {action.complete && action.slug === 'cozy-collect' &&
```
gregorylegarec: Same remark

gregorylegarec:
Thanks @CPatchane !

PR #69, opened 2017-09-14
Specify the jobs permissions

jinroh:
Specify the jobs permissions to explicitly ask for the `sendmail` worker. This evolution is required by this change in cozy-stack: https://github.com/cozy/cozy-stack/pull/950


gregorylegarec:
Thanks @jinroh :beers:

PR #70, opened 2017-09-25
Remove product board email sending

CPatchane:


CPatchane:
[(1) [NUA-9] Changer la redirection du nuage](https://trello.com/c/OjC3S93B/616-1-nua-9-changer-la-redirection-du-nuage)

PR #71, opened 2017-09-26
NUA-8: improve alert message when sending email to support

m4dz:


src/services/Claudy.jsx
```diff
@@ -205,6 +210,7 @@ export class Claudy extends Component {
                 iconSrc={this.getIcon(selectedAction.icon)}
                 url={selectedActionUrl}
                 onActionClick={() => this.trackActionLink(selectedAction)}
+                onSuccess={this.goBack}
```
gregorylegarec: I think we should ensure that we are using the right context, even if it seems to work as it.
```js
    onSuccess={() => this.goBack()}
```

src/services/ClaudyActionComponents/Support.jsx
```diff
@@ -22,9 +22,11 @@ export class Support extends Component {
       nextProps.opened ? this.onOpen() : this.onReturn()
     }
 
-    // reset message if successfully sent
     if (nextProps.emailStatus.isSent && this.props.emailStatus.isSending) {
+      // reset message if successfully sent
```
gregorylegarec: I feel that the code is enough self-explainatory to avoid those two comments. What do you think ?

src/services/Claudy.jsx
```diff
@@ -161,10 +161,13 @@ export class Claudy extends Component {
           <button className='coz-btn-close' onClick={onClose} />
           <button className='coz-claudy-menu-header-back-button' onClick={this.goBack} />
         </header>
-        {!emailStatus.isSending && emailStatus.isSent &&
-          <div className='coz-claudy-menu-action-description-success'>
-            {t('claudy.actions.support.success')}
-          </div>
+        { // FIXME: We use a `p` tag as long as a `div` will break the support
```
gregorylegarec: Ping @goldoraf, we all saw this by our own eyes. Do you have a beginning of an explanation ?

src/services/Claudy.jsx
m4dz: Hmm, agree, but I need to pass an optional argument, see https://github.com/cozy/cozy-settings/pull/71/files#diff-262290da52d26b12b91bcea1395c10e6R125 and https://github.com/cozy/cozy-settings/pull/71/files#diff-dad91f6e3ef7c4e1710f2ad844a0f6e7R29.

So maybe we could just bind the function to the context: `onSuccess={this.goBack.bind(this)}`, so we keep the context and we can pass args if needed. Is it better?

src/services/ClaudyActionComponents/Support.jsx
m4dz: Hmm… Agree, finally the line is more to explain the `successfully sent` rather than the `reset message` 😆

src/services/Claudy.jsx
CPatchane: and what about `{(alert) => this.goBack(alert)}` ?

gregorylegarec: Actually, I personnally try to avoid binding like this, I feel them confusing.
Furthermore, @goldoraf told me back in the days that the notation `foo={() => bar()}` was more "React". I'd love to have him confirming this point.

goldoraf: Weird that it works this way, without binding `this`... `foo={() => bar()}` is a great way to avoid binding this at a small performance cost, another way is using [class properties](https://babeljs.io/docs/plugins/transform-class-properties/).

enguerran: Please read this little article (~ 3 minutes) about arrow functions or bind in props by Cory House : https://medium.freecodecamp.org/why-arrow-functions-and-bind-in-reacts-render-are-problematic-f1c08b060e36

gregorylegarec: Isn't the case presented in the article only related to having an arrow function into a `map` into a `render` ?
I think that we are not concerned here, but it is still really interesting to keep in mind what is explained in the article.

enguerran: Lorsque tu passes une arrow function ou que tu bind une fonction dans la props, le composant parent P passe à chaque fois qu'il est rendu une nouvelle instance de cette props à ses enfants.
Ainsi `PureComponent` et `shouldComponentUpdate` seront cassés car ce seront des nouvelles instances, donc pas de comparaison possible pour le VDOM.

> The parent component is passing down an arrow function on props. Arrow functions are reallocated on every render (same story with using bind). So although I’ve declared User.js as a PureComponent, the arrow function in User’s parent causes the User component to see a new function being sent in on props for all users. So every user re-renders when any delete button is clicked. 👎

Si tu passes une même instance à la props des enfants, alors cette instance pourra être comparée et re-rendu seulement si nécessaire.

Regarde le deuxième exemple de l'article. Il utilise une class properties pour déclarer ses composants enfants (permettant le scoping de this) et on voit bien dans la console que les enfants ne sont pas re-rendus à chaque delete: https://codesandbox.io/s/jnowr0ww7v

enguerran: Attention : mon message n'est pas un appel à faire autrement, c'est juste pour informer de l'inconvénient de arrow/bind dans les props. Les problèmes d'optimisation viendront en temps voulu.

gregorylegarec:
Thanks @m4dz 💃

PR #78, opened 2017-10-10
feat: connection historic

Narkfr:


Narkfr:
[(5) [SET-21] ETU j'ai accès à l'historique des connexions](https://trello.com/c/W4PxruTZ/630-5-set-21-etu-jai-acc%C3%A8s-%C3%A0-lhistorique-des-connexions)

src/components/SessionsView.jsx
```diff
@@ -0,0 +1,62 @@
...
+                        new Date(session.created_at),
+                        t('SessionsView.sync_date_format')
+                      )
+                    }
+                  </div>
+                  <div className={classNames(tableStyles['coz-table-cell'], tableStyles['set-table-os'])}>
+                    {session.os}
+                  </div>
+                  <div className={classNames(tableStyles['coz-table-cell'], tableStyles['set-table-browser'])}>
+                    {parser(session.user_agent).browser.name} {parser(session.user_agent).browser.major}
```
CPatchane: Maybe parsing the `session.user_agent` only once before the `render` function could be better, what do you think?

Narkfr: session is defined in the map function, I can't access it in the render. 
But I can make a sub component <SessionViewRow session={session} /> :+1:

src/components/SessionsView.jsx
```diff
@@ -0,0 +1,62 @@
...
+import React from 'react'
+import classNames from 'classnames'
+import parser from 'user-agent-parser'
+
+import Loading from './Loading'
+
+import { translate } from 'cozy-ui/react/helpers/i18n'
+
+const SessionsView = ({ t, f, isFetching, sessions, deleteOtherSessions }) => {
+  return (
```
CPatchane: You can put just above here

```
const user_agent = parser(session.user_agent)
```

can't you?

CPatchane:
Thanks @Narkfr 🎉

PR #87, opened 2017-11-06
feat: restore delete sessions button

Narkfr:


Narkfr:
[(3) [SET-14]  ETU je peux déconnecter mes sessions courantes](https://trello.com/c/7ohNnHmK/662-3-set-14-etu-je-peux-d%C3%A9connecter-mes-sessions-courantes)

src/actions/index.js
```diff
@@ -177,22 +177,39 @@ export const deviceModaleRevokeClose = () => ({
...
         alert: {
           message: 'SessionsView.infos.server_error'
         }
       })
+    }
+
+    // Merge ID and Sessions to inject only currents sessions in the store
+    const currentsSessions = []
+
+    responseCurrentsSessionsIds.data.map((currentSessionId) => {
```
Narkfr: I can probably use a shiny ES6 function here with .reduce or whatever, if you have it in mind !

CPatchane: I suggest:

```js
currentsSessions.filter(
  session =>responseCurrentsSessionsIds.data.includes(session.session_id)
)
```

Narkfr: :sparkles: :sparkles: Thx :)

Narkfr: .includes doesn't work with an array of object :/ my solution isn't the more elegant, but it works, lets merge it ?

CPatchane: Oh I didn't notice the `.id`, is it better? 

```
const currentSessions = sessions.filter( session => 
  for (let session of sessions) {
    return currentSessionId.id === session.session_id
  }
)
```

CPatchane: We can keep it like, it's correct too 👍

PR #97, opened 2018-01-15
fix: table was broken on mobile views :bug:

CPatchane:


CPatchane:
[(5) ETD j'ai migré sur la nouvelle version de cozy-ui [2]](https://trello.com/c/xa695nXd/754-5-etd-jai-migr%C3%A9-sur-la-nouvelle-version-de-cozy-ui-2)

src/styles/table.styl
```diff
@@ -32,9 +32,13 @@ $set-device
         flex 0 0 15%
 
     .set-table-devices
+        height auto!important //@stylint ignore
```
gregorylegarec: We must not use `!important`. Why can't we avoid using it here ?

CPatchane: Actually this is a quickfix for cozy-ui, I forget to put a comment here but that will be removed as soon as tables in cozy-ui are ready.

gregorylegarec:
Thanks @CPatchane, we must not forget to update this code as soon as Cozy UI is fixed. :)

PR #100, opened 2018-01-17
Add offers button if offers link are available

CPatchane:
The offers link is compute from an URL in the context and the UUID of the instance. If one of these parameters are missing or `null` this part won't be displayed in the view.
screen shot 2018-01-17 at 10 28 31

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -60,8 +60,30 @@ export const fetchInfos = () => {
 }
 
 export const fetchStorageData = () => {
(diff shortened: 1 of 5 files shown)


CPatchane:
[(3) ETU, je peux faire une demande pour avoir un premium](https://trello.com/c/nA71adw5/800-3-etu-je-peux-faire-une-demande-pour-avoir-un-premium)

gregorylegarec:
Thanks @CPatchane 🍨

PR #110, opened 2018-01-31
feat: enable 2fa

Narkfr:


Narkfr:
[(5) ETU je peux activer ma double authentification](https://trello.com/c/i2ztjOD9/781-5-etu-je-peux-activer-ma-double-authentification)

PR #111, opened 2018-02-06
feat: real 2fa activation

Narkfr:


src/components/ProfileView.jsx
```diff
@@ -314,6 +345,11 @@ const MailConfirmationCode = translate()(({
         <p>{t('ProfileView.twofa.modal.nocode_claude')}<a href='mailto:claude@cozycloud.cc'>claude@cozycloud.cc</a></p>
       </div>
     </div>
+    {
+      fields.email.value === '' && <p className={styles['coz-form-errors']}>
```
gregorylegarec: Does ` fields.email.value` can be `null` ? Does it still supposed to work ?

src/components/ProfileView.jsx
```diff
@@ -323,11 +359,12 @@ const MailConfirmationCode = translate()(({
       </Button>
       <Button
         onClick={() => checkMailConfirmationCode(fields.mail_confirmation_code.value)}
+        disabled={fields.email.value === ''}
```
gregorylegarec: Same question here. As `!''` is `true`, we could only check for falsy/truthy value by just checking `fields.email.value`. It could also solve cases when `fields.email.value` is `null` or `undefined`.

src/components/ProfileView.jsx
Narkfr: value is set by default to '' by the reducer https://github.com/cozy/cozy-settings/blob/master/src/reducers/fields.js#L12 I'm not really palcid and enthusiastic to change this feature

gregorylegarec: Yep, but if we only consider falsy value, it will continue to work and will also be future-proof. But if you not convinced, let's go with it (It is a concern that should be discussed and statuated by our  - unexisting - style guide).

gregorylegarec:
Thanks @Narkfr 🌵

PR #125, opened 2018-05-14
Update to cozy-ui v7.22.0

gregorylegarec:


src/actions/passphrase.js
```diff
@@ -14,14 +14,15 @@ export const UPDATE_PASSPHRASE_2FA_2 = 'UPDATE_PASSPHRASE_2FA_2'
...
-export const updatePassphrase = (current, newVal) => {
+export const updatePassphrase = (current, newVal, callback) => {
   return async (dispatch, getState) => {
     dispatch({ type: UPDATE_PASSPHRASE })
     return cozyFetch('PUT', '/settings/passphrase', {
       'current_passphrase': current,
       'new_passphrase': newVal
     }).then(instance => {
       dispatch({ type: UPDATE_PASSPHRASE_SUCCESS })
+      callback()
```
gregorylegarec: Why using a callback as we are already in an asynchronous context, which the role is specially to avoid callbacks ?

src/containers/Profile.jsx
```diff
@@ -42,17 +42,22 @@ const mapDispatchToProps = (dispatch, ownProps) => ({
     dispatch(desactivate2FA(mode))
   },
   onPassphraseSimpleSubmit: (current, newVal) => {
-    return dispatch(updatePassphrase(current, newVal)).catch(
```
gregorylegarec: Ok now I see why it did not work. Replacing the `catch` by a `then` should make it.

CPatchane: No, we have to call the alerter before the setTimeout so before the promise return. It's why I used a callback here because `then` was not working correctly.

gregorylegarec: I need to check by myself because to me the `setTimeout` is asynchronous, and we already fulfilled the promise when it is called.

PR #132, opened 2018-08-27
Add gather action

gregorylegarec:
The gather action is the same as cozy-collect, so this PR rename the variables related to collect, but keeps also the cozy-collect action to ensure legacy. The cozy-collect action should be removed soon, once the Cozy-Bar will be updated.


src/services/Claudy.jsx
```diff
@@ -83,7 +84,8 @@ export class Claudy extends Component {
           action.complete = !!claudyInfos.devices.find(d => d.client_kind === MOBILE_CLIENT_KIND)
           break
         case 'cozy-collect':
-          action.complete = !!claudyInfos.collectAccounts.length
+        case 'gather':
```
CPatchane: Maybe use `CLAUDY_ACTION_GATHER` constant here?

PR #168, opened 2019-01-29
Better conflict handling with retries

gregorylegarec:
When an update on instance returns a 409, make 3 more attempt by retrieving the remote instance and retrying the update.


src/actions/index.js
```diff
@@ -147,27 +166,35 @@ export const updateInfo = (field, value) => {
     }
     // tracking field must be stored as string
     if (field === 'tracking') value = value.toString()
-    let newInstance = Object.assign({}, getState().instance)
+    const instance = getState().instance
+    const newInstance = Object.assign({}, instance)
```
nono: I'm not sure to see what you are doing here. Are you trying to clone `instance` to avoid modifying the old version?

src/actions/index.js
```diff
@@ -125,8 +125,27 @@ export const fetchStorageData = () => {
   }
 }
 
+const tryUpdate = async (instance, { retries = 0 }) => {
+  let updatedInstance
+  try {
+    updatedInstance = await cozyFetch('PUT', '/settings/instance', instance)
```
nono: [nitpick] you can put `return await cozyFetch('PUT', '/settings/instance', instance)` here

src/actions/index.js
gregorylegarec: Yes, I kept this part as it already was. But
```js
const instance = { ...getState().instance }
```
is actually a better way, I will update this part.

nono: My point was that it doesn't work with the next line:

```js
const field = "foo"
const value = "baz"
const oldInstance = { data: { foo: "bar" } }
const instance = { ...instance } // or const instance = Object.assign({}, oldInstance)
instance.data.attributes[field] = value
console.log(oldInstance.data.foo) // => it returns "baz", not "bar"
```

You have to used deep cloning if what you want is to prevent modification of the old version.

gregorylegarec: You're totally right.

PR #170, opened 2019-02-05
Get twoFA property based on auth_mode settings

CPatchane:
Don't use a custom two_fa property anymore

### DIFF
--- a/src/actions/twoFactor.js
+++ b/src/actions/twoFactor.js
@@ -1,4 +1,4 @@
-import { cozyFetch, updateInfo } from './index'
+import { cozyFetch } from './index'
 
 export const ACTIVATE_2FA = 'ACTIVATE_2FA'
 export const ACTIVATE_2FA_FAILURE = 'ACTIVATE_2FA_FAILURE'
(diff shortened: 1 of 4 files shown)


CPatchane:
[[Settings] 2FA marqué comme désactivé dans settings alors que c'est actif côté stack](https://trello.com/c/1vtVWiN5/927-settings-2fa-marqu%C3%A9-comme-d%C3%A9sactiv%C3%A9-dans-settings-alors-que-cest-actif-c%C3%B4t%C3%A9-stack)

nono:
Ping @gregorylegarec

src/actions/twoFactor.js
```diff
@@ -75,8 +75,11 @@ export const checkTwoFactorCode = (code, mode = AUTH_MODE.TWO_FA_MAIL) => {
       auth_mode: mode
     })
       .then(() => {
-        dispatch({ type: CHECK_TWO_FACTOR_CODE_SUCCESS })
-        dispatch(updateInfo('two_fa', true))
+        dispatch({
+          type: CHECK_TWO_FACTOR_CODE_SUCCESS,
+          field: 'auth_mode',
```
gregorylegarec: Is `auth_mode` a new field ? If so I think we should use `authMode` instead to be compliant with recent guidelines.

CPatchane: This is a field based directly on the stack name, it's why this is not camelCase here

PR #173, opened 2019-02-07
Better handling errors from profile actions

CPatchane:
Fix the exact same issue than for #172 but for other profile actions

### DIFF
--- a/src/containers/Profile.jsx
+++ b/src/containers/Profile.jsx
@@ -37,8 +37,12 @@ const mapDispatchToProps = (dispatch, ownProps) => ({
     dispatch(updateInfo(field, value))
   },
   requestExport: async () => {
-    await dispatch(requestExport())
-    Alerter.success(ownProps.t('ProfileView.export.success'))
+    try {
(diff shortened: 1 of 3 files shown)


CPatchane:
[[SET] Message de confirmation même en cas d'erreur de mot de passe](https://trello.com/c/PjdliAmY/903-set-message-de-confirmation-m%C3%AAme-en-cas-derreur-de-mot-de-passe)

src/containers/Profile.jsx
```diff
@@ -37,7 +37,11 @@ const mapDispatchToProps = (dispatch, ownProps) => ({
     dispatch(updateInfo(field, value))
   },
   requestExport: async () => {
-    await dispatch(requestExport())
+    try {
+      await dispatch(requestExport())
```
nono: You should move the line `Alerter.success(ownProps.t('ProfileView.export.success'))` here, to avoid showing the success alerter when there is an error

CPatchane: Oh, forgot it. Thanks 👍

PR #185, opened 2019-02-19
Do not freeze the browser on support email sent

nono:
In the claudy intent, the user can send an email to the support. When it's done, claudy comes back to the list of actions. But the componentWillReceiveProps method of Support was calling the goBack method of Claudy, when it makes React re-rendering and calling again the first method, etc. To fix that, we have added a condition to break the loop. It is a quick fix, a proper refactoring would need more work but could clean a lot more.


gregorylegarec:
Dealing with this kind of logic into `componentWillReceiveProps` is always painful and seems to be a bad pattern. Thanks @nono !

PR #186, opened 2019-02-20
Improve html

nono:
- Fix the glitch when the app is loading
- Use the favicon helper


nono:
[ETU j'ai quelque fois un gros nuage au chargement de mes applications](https://trello.com/c/yRCkm1Dz/742-etu-jai-quelque-fois-un-gros-nuage-au-chargement-de-mes-applications)

src/targets/browser/index.ejs
```diff
@@ -1,5 +1,6 @@
 <!DOCTYPE html>
 <html lang="{{.Locale}}">
+<head>
```
gregorylegarec: m4dz point was that browsers are smart enought to fix this point by themself.

I a pretty much no opinion so this version fits me as well.

gregorylegarec:
Thanks @nono 👍

src/targets/browser/index.ejs
nono: I confirm: the browsers are smart enough to do that.

But this trick was not used consistently: `<head>` was still present in a lot of templates in our applications, and I prefer to be explicit. So, I've put again `<head>` in our templates, even if it is not strictly needed.

PR #193, opened 2019-03-06
fix: Use ua-parser-js for reading user-agents

### LINKED ISSUE #154: Dependency deprecation warning: user-agent-parser (npm)
On registry `https://registry.npmjs.org/`, the "latest" version (v0.6.0) of dependency `user-agent-parser` has the following deprecation notice:

`use ua-parser-js instead`

Marking the latest version of an npm package as deprecated results in the entire package being considered deprecated, so contact the package author you think this is a mistake.

Affected package file(s): `package.json`

If you don't care about this, you can close this issue and not be warned about `user-agent-parser`'s deprecation again. If you would like to completely disable all future deprecation warnings then add the following to your config:

```
"suppressNotifications": ["deprecationWarningIssues"]
```

nono:
The user-agent-parser package on npm has been deprecated in favor of ua-parser-js. Let's use the modern alternative, with support for the recent browsers and OS.

Fix #154

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,6 @@
     "redux": "4.0.1",
     "redux-logger": "3.0.6",
     "redux-thunk": "2.3.0",
-    "user-agent-parser": "0.6.0"
+    "ua-parser-js": "0.7.19"
   }
 }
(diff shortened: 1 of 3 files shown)


nono:
[[SET] Remplacer user-agent-parser qui  n'est plus maintenue](https://trello.com/c/P54vcg8m/1084-set-remplacer-user-agent-parser-qui-nest-plus-maintenue)

gregorylegarec:
Thanks @nono 👍

PR #209, opened 2019-04-08
Use React + fix some react warnings

CPatchane:
Stop using preact in production, use React instead
Also use Select component from Cozy-UI for better consistent


CPatchane:
[🐌 [SET] Migration preact/React](https://trello.com/c/ykAE3CpP/1215-%F0%9F%90%8C-set-migration-preact-react)

src/components/PassphraseForm.jsx
```diff
@@ -62,7 +62,7 @@ class PassphraseForm extends Component {
           key="current_password"
           value={currentPassword}
           inError={currentPasswordError}
-          onInput={e => this.handleCurrentInput(e)}
+          onChange={e => this.handleCurrentInput(e)}
```
gregorylegarec: Why removing `onInput` ? `onInput` and `onChange` does not have the same behaviour.

PR #216, opened 2019-04-26
Use transpiled cozy-ui

gregorylegarec:
This PR applies guidelines from https://github.com/cozy/cozy-ui/blob/master/docs/how-to-transpile.md#how-to

### DIFF
--- a/app.config.js
+++ b/app.config.js
@@ -1,7 +1,8 @@
 const configs = [
   require('cozy-scripts/config/webpack.bundle.default'),
   require('cozy-scripts/config/webpack.config.css-modules'),
-  require('./config/webpack.config.piwik')
+  require('./config/webpack.config.piwik'),
+  require('./config/webpack.config.cozy-ui')
 ]
 
 module.exports = configs
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -2,8 +2,10 @@ import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
 import { hot } from 'react-hot-loader'
+
+import 'cozy-ui/react/stylesheet.css'
+import { IconSprite } from 'cozy-ui/react'
 import { translate } from 'cozy-ui/react/I18n'
-import { Sprite as IconSprite } from 'cozy-ui/react/Icon'
 import { Layout, Main } from 'cozy-ui/react/Layout'
 import { Route, Switch, Redirect, withRouter } from 'react-router-dom'
 
--- a/config/webpack.config.cozy-ui.js
+++ b/config/webpack.config.cozy-ui.js
@@ -1,0 +1,9 @@
+'use strict'
+
+module.exports = {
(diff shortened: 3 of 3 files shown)


app.config.js
```diff
@@ -1,7 +1,8 @@
 const configs = [
   require('cozy-scripts/config/webpack.bundle.default'),
   require('cozy-scripts/config/webpack.config.css-modules'),
-  require('./config/webpack.config.piwik')
+  require('./config/webpack.config.piwik'),
+  require('./config/webpack.config.cozy-ui')
 ]
```
CPatchane: Why not using an alias here instead of adding `/transpiled` everywhere?

app.config.js
```diff
@@ -1,7 +1,8 @@
 const configs = [
   require('cozy-scripts/config/webpack.bundle.default'),
   require('cozy-scripts/config/webpack.config.css-modules'),
-  require('./config/webpack.config.piwik')
+  require('./config/webpack.config.piwik'),
+  require('./config/webpack.config.cozy-ui')
```
edas: This file seems missing in the PR

app.config.js
edas: If an alias is working, that's also fine with me a easy to test/remove later

app.config.js
gregorylegarec: Classic, thanks !

src/components/App.jsx
```diff
@@ -2,13 +2,15 @@ import React, { Component } from 'react'
...
+
+import 'cozy-ui/transpiled/react/stylesheet.css'
+import { IconSprite } from 'cozy-ui/transpiled/react'
+import { translate } from 'cozy-ui/transpiled/react/I18n'
+import { Layout, Main } from 'cozy-ui/transpiled/react/Layout'
 import { Route, Switch, Redirect, withRouter } from 'react-router-dom'
 
 import Sidebar from 'components/Sidebar'
-import Alerter from 'cozy-ui/react/Alerter'
+import Alerter from 'cozy-ui/transpiled/react/Alerter'
```
edas: Did you forgot to update this file and remove the "transpiled" in it ?

gregorylegarec: Probably. We cannot trust Search and replace :(

gregorylegarec: Thanks !

src/components/App.jsx
```diff
@@ -2,8 +2,10 @@ import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
 import { hot } from 'react-hot-loader'
+
+import 'cozy-ui/react/stylesheet.css'
+import { IconSprite } from 'cozy-ui/react'
```
edas: Je suis gêné par cette modif. Au minimum elle est indépendante de cette PR, non ?

gregorylegarec: C'est l'application du point 3.

PR #227, opened 2019-05-14
fix: Update checkbox after deactivating 2FA

ptbrowne:
Reducer did not take into account the action DESACTIVATE_2FA_SUCCESS.

### DIFF
--- a/src/reducers/fields.spec.js
+++ b/src/reducers/fields.spec.js
@@ -1,0 +1,61 @@
+import fields from './fields'
+import {
+  CHECK_TWO_FACTOR_CODE_SUCCESS,
+  DESACTIVATE_2FA_SUCCESS,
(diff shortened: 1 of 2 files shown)


src/reducers/fields.spec.js
```diff
@@ -0,0 +1,61 @@
...
+    state = fields(state, {
+      type: CHECK_TWO_FACTOR_CODE_SUCCESS
+    })
+    expect(state.auth_mode.value).toBe(AUTH_MODE.TWO_FA_MAIL)
+    state = fields(state, {
+      type: DESACTIVATE_2FA_SUCCESS
+    })
+    expect(state.auth_mode.value).toBe(AUTH_MODE.BASIC)
+  })
+})
```
CPatchane: 👍

PR #243, opened 2019-06-06
Fix delete account

CPatchane:
We don't need to have context here for the domain any more and it wasn't working since a while.

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -1,5 +1,4 @@
 import React, { Component } from 'react'
-import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
 import { hot } from 'react-hot-loader'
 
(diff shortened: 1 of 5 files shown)


CPatchane:
[🔵 🍊 [SET] Perte du domain à l'envoie de la demande de suppression](https://trello.com/c/qBYZRjgN/1485-%F0%9F%94%B5-%F0%9F%8D%8A-set-perte-du-domain-%C3%A0-lenvoie-de-la-demande-de-suppression)

edas:
Could you add a test ?

CPatchane:
I changed to make the FormModal reading directly the domain from the HTML page and remove the props passing here. Do you see a good way to test this case efficiently @edas?

edas:
> I changed to make the FormModal reading directly the domain from the HTML page and remove the props passing here. Do you see a good way to test this case efficiently @edas?

You can 
- inject a fake HTML document with jest
- mock `sendDeleteAccountRequest`

Then 
- check that sendDeleteAccountRequest subject does contain the domain
- check that sendDeleteAccountRequest subject does not contain %{domain}

PR #269, opened 2019-12-12
Passphrase update extract and new UI

drazik:
This introduces three things about the passphrase update form:

* Extract it to its own page
* Update the UI (with hint update)
* Handle `redirect_success` and `redirect_cancel` URL params to be able to redirect to another app on success or cancel. We need this because we want to redirect to the password manager onboarding app when we come from this app to update the passphrase

Some code changes are not directly related to the main subject of the PR but are related to cozy-ui and cozy-client update. Cozy-ui update was needed to get `InputGroup` component. But updating cozy-ui required a cozy-client update.

I also did some cleaning on naming because we had passphrase and password terms mixed which annoyed me when reading the code.

When the PR is merged, we still need to add new translations for french at least.

image


src/actions/passphrase.js
```diff
@@ -163,15 +168,33 @@ export const updatePassphrase2FASecond = (
...
       })
   }
 }
+
+export const updateHint = hint => {
+  return async dispatch => {
+    dispatch({ type: UPDATE_HINT })
+
+    try {
+      await cozyFetch('PUT', '/settings/hint', { hint })
```
Crash--: is it too complicated to start to use cozy-client here?

src/components/PassphraseExample.jsx
```diff
@@ -0,0 +1,70 @@
...
+  if (isAlphaCharacter(char)) {
+    return tokenTypes.alpha
+  }
+
+  if (isNumberCharacter(char)) {
+    return tokenTypes.number
+  }
+
+  return tokenTypes.special
+}
```
Crash--: Why can we have several types of token?

src/components/2FA/Passphrase2FA.jsx
```diff
@@ -1,12 +1,12 @@
 import React, { Component } from 'react'
 import { translate } from 'cozy-ui/react/I18n'
 import Modal, { ModalContent } from 'cozy-ui/react/Modal'
+import Input from 'cozy-ui/react/Input'
```
Crash--: Do settings has an alias somewhere to use the transpiled version?

src/components/PassphraseExample.jsx
```diff
@@ -0,0 +1,70 @@
...
+
+const tokenize = passphrase => {
+  const tokens = passphrase.split('').map(char => ({
+    char,
+    type: getTokenType(char)
+  }))
+
+  return tokens
+}
```
Crash--: I don't understand the manipulation we're doing on the passphrase. Can you explain why we have to map on each char to check the Type?

Crash--: Oh, is it to check the strength of the password?

src/components/PassphraseForm.jsx
```diff
@@ -1,123 +1,225 @@
...
+                'PassphraseView.new_passphrase.confirmation_placeholder'
+              )}
+              value={newPassphraseRepeat}
+              onChange={this.handleInputChange}
+              error={newPassphraseTouched && !newPassphraseMatch}
+            />
+          </Stack>
+          {newPassphraseTouched &&
+            !newPassphraseMatch && (
+              <p className="u-error">
```
edas: Pointilleux: Un composant ErrorMessage ou quelque chose du genre plutôt que pointer une classe ad-hoc sur un paragraphe ?

src/components/PassphraseExample.jsx
drazik: Because we want to color alpha, numeric and others (considered as special) characters in the password example differently:

image

So I cut the string and give a type to each char

src/components/2FA/Passphrase2FA.jsx
drazik: Yes there's one ;)

src/components/PassphraseExample.jsx
drazik: See https://github.com/cozy/cozy-settings/pull/269#discussion_r357227962

src/components/PassphraseForm.jsx
```diff
@@ -1,123 +1,225 @@
...
+              </p>
+            )}
+          {globalError && <p className="u-error">{t(globalError)}</p>}
+          {twoFactorError && <p className="u-error">{t(twoFactorError)}</p>}
+          <Stack
+            spacing="m"
+            tag="ul"
+            className={styles['coz-passphrase-advices']}
+          >
+            <Text tag="li">
```
edas: Ça me gêne un peu d'avoir une fausse liste en utilisant les attributs tag. Pourquoi ne pas avoir utilisé une vraie liste ? 

```
<Stack>
  <List>
   <ListItem>
     <Text>
```

src/components/PassphraseExample.jsx
Crash--: Got it! Thanks. It makes more sense now :)

src/components/PassphraseExample.jsx
```diff
@@ -0,0 +1,70 @@
...
+    </span>
+  )
+}
+
+const PassphraseExample = props => {
+  const { passphrase, ...rest } = props
+  const tokens = tokenize(passphrase)
+
+  return (
+    <span className={styles['coz-passphrase-example']} {...rest}>
```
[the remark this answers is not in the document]
drazik: Ping @joel-costa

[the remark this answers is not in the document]
drazik: Vu avec @joel-costa : oui on extrait ça dans cozy-ui

src/components/PassphraseForm.jsx
drazik: Et on met ErrorMessage dans cozy-ui du coup, non ?

src/components/Input.jsx
```diff
@@ -47,54 +52,68 @@ export default translate()(props => (
...
+    >
+      <Icon
+        icon={hidden ? 'eye' : 'eye-closed'}
+        size={16}
+        color="var(--coolGrey)"
+      />
+    </button>
+  )
 }
+export const PassphraseInput = props => {
```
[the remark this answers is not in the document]
drazik: Après avoir regardé le Field de cozy-ui avec le type "password", je me suis dis que j'allais d'abord implémenter cette nouvelle forme séparément, pour avancer plus vite car ça me paraissait nécessiter pas mal de modifications sur le Field.

J'ai créé https://github.com/cozy/cozy-ui/issues/1304

src/components/PassphraseExample.jsx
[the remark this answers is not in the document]
drazik: See https://github.com/cozy/cozy-ui/pull/1303

src/components/PassphraseForm.jsx
drazik: J'ai extrait le style dans un composant List / ListItem dans cozy-ui. On pourra utiliser ça à la place comme ça : https://github.com/cozy/cozy-ui/pull/1305

src/components/PassphraseForm.jsx
edas: yep !

src/components/Input.jsx
```diff
@@ -47,54 +52,68 @@ export default translate()(props => (
...
-          placeholder={t(`ProfileView.${name}.placeholder`)}
-          value={value}
-          onChange={onChange}
-          className={inError ? styles['error'] : ''}
-          autoComplete={autocomplete || 'off'}
-        />
-      </div>
-    )
-  }
+const HideShowButton = props => {
```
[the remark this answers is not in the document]
drazik: Ça va avec https://github.com/cozy/cozy-settings/pull/269#discussion_r357225768

src/actions/passphrase.js
drazik: I already consumed a lot of time on this PR, I'd rather not touch this part for now.

src/components/PassphraseView.jsx
```diff
@@ -0,0 +1,71 @@
...
+    onPassphrase2FAStep2,
+    passphrase,
+    isTwoFactorEnabled,
+    instance,
+    fetchInfos
+  } = props
+
+  useEffect(() => {
+    fetchInfos()
+  }, [])
```
ptbrowne: It is a bit surprising to have the component call fetchInfos but to see not update of state in this component. It seems like the parent is holding the "infos" but in this case, why is not the parent that does fetchInfos ?

src/components/PassphraseView.jsx
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  const handlePassphrase2FASubmit = twoFactorCode => {
+    const { twoFactorToken } = passphrase
+    onPassphrase2FAStep2(
+      currentPassphrase,
+      new2FAPassphrase,
+      twoFactorCode,
+      twoFactorToken
+    ).then(() => setTwoFAModalOpen(false))
```
ptbrowne: It seems the UI is reacting after a network round trip but no visual indication is given to the user, I see no setPassphrase2FAStepSubmitting ? Is it expected ?

src/containers/Passphrase.jsx
```diff
@@ -0,0 +1,64 @@
...
+  fetchInfos: () => dispatch(fetchInfos())
+})
+
+const Password = compose(
+  translate(),
+  connect(
+    mapStateToProps,
+    mapDispatchToProps
+  )
+)(PassphraseView)
```
ptbrowne: Ah OK, I understand better my question on the fetchInfos from the parent. I think it would be clearer if all the dispatch functions and the connect were in the PassphraseView component.

src/containers/Passphrase.jsx
```diff
@@ -0,0 +1,64 @@
...
+          current,
+          newVal,
+          twoFactorCode,
+          twoFactorToken
+        )
+      )
+        .then(() => Alerter.info(ownProps.t('ProfileView.password.reload')))
+        // eslint-disable-next-line no-console
+        .catch(e => console.error(e))
+    )
```
ptbrowne: Is there a reason not to use await/async ?

package.json
```diff
@@ -46,11 +46,12 @@
   },
   "dependencies": {
     "classnames": "2.2.6",
-    "cozy-client": "6.48.0",
+    "cozy-client": "8.7.1",
```
ptbrowne: > I also had to add cozy-doctypes dep to not
get errors

Do you think it is a bug that needs solving in cozy-client ?

src/components/Input.jsx
```diff
@@ -97,4 +102,70 @@ class PasswordInputComponent extends Component {
   }
 }
 
+const PasswordStrength = props => {
+  const { password, className, ...rest } = props
+  const strength = passwordHelper.getStrength(password)
```
ptbrowne: Nit: If getStrength is expensive (since it uses regexes), a useMemo could be useful.

src/components/Input.jsx
```diff
@@ -97,4 +102,70 @@ class PasswordInputComponent extends Component {
...
+  const [hidden, setHidden] = useState(true)
+
+  return (
+    <div className={cx(styles['coz-pwd-input'], className)}>
+      <InputGroup
+        append={
+          <HideShowButton hidden={hidden} onClick={() => setHidden(!hidden)} />
+        }
+        className={cx(
+          styles['coz-input-group'],
```
ptbrowne: Nit: This styles removes the border of the `input`, I think it would be best if this behavior was coded inside InputGroup.

src/containers/Passphrase.jsx
```diff
@@ -21,9 +22,12 @@ const mapStateToProps = state => ({
...
 
 const mapDispatchToProps = (dispatch, ownProps) => ({
-  onPassphraseSimpleSubmit: (current, newVal) => {
+  onPassphraseSimpleSubmit: (currentPassphrase, newPassphrase, hint) => {
     return (
-      dispatch(updatePassphrase(current, newVal))
+      dispatch(updateHint(hint))
+        .then(() =>
+          dispatch(updatePassphrase(currentPassphrase, newPassphrase))
+        )
```
ptbrowne: I think using await would be more readable.

src/reducers/passphrase.js
```diff
@@ -70,10 +76,12 @@ const errors = (state = null, action) => {
     case UPDATE_PASSPHRASE_2FA_1:
     case UPDATE_PASSPHRASE_SUCCESS:
     case UPDATE_PASSPHRASE_2FA_2_SUCCESS:
+    case UPDATE_HINT_SUCCESS:
       return null
     case UPDATE_PASSPHRASE_2FA_1_FAILURE:
     case UPDATE_PASSPHRASE_2FA_2_FAILURE:
     case UPDATE_PASSPHRASE_FAILURE:
+    case UPDATE_HINT_FAILURE:
```
ptbrowne: This code seems too complex from what it is doing. Everytime we add something in the form we have to add stuff here :/ IMHO, redux should not be used here as the state is not shared between a lot of components or between components that do not share a common parent. This is not in the scope of the PR.

src/containers/Passphrase.jsx
```diff
@@ -21,14 +21,24 @@ const mapStateToProps = state => ({
...
 })
 
+const showSuccessThenReload = t => {
+  Alerter.info(t('PassphraseView.reload'))
+
+  setTimeout(() => {
+    // the token changes after a password change, so we need to reload
+    // the page to get the new one
+    window.location.reload()
+  }, 4000) // 4s, a bit longer than the alert message
```
ptbrowne: Setting the duration explicitly in Alerter.info would allow us to remove the comment.

```
const ALERT_DURATION = 3500
   Alerter.info(t('PassphraseView.reload'), { duration: ALERT_DURATION })

setTimeout(() => {...}, ALERT_DURATION + 500) 

```

src/containers/Passphrase.jsx
```diff
@@ -21,13 +22,32 @@ const mapStateToProps = state => ({
   passphrase: state.passphrase
 })
 
-const showSuccessThenReload = t => {
-  Alerter.info(t('PassphraseView.reload'))
+export const getRedirectUrlsFromURLParams = urlParamsStr => {
```
ptbrowne: Nit: parse would be explicit about the need to parse the URL.

src/actions/passphrase.js
```diff
@@ -25,9 +25,11 @@ const getInstanceURL = () => {
   return client.getStackClient().uri
 }
 
-const invalidPasswordErrorAction = {
+const invalidPassphraseErrorAction = {
```
ptbrowne: Kudos on doing the rename from passphrase to password !

src/components/Input.jsx
drazik: This has been extracted to cozy-ui, in which the regex have been extracted from the helper function, so it's not recreated on every call.

src/components/Input.jsx
drazik: This has been extracted to cozy-ui (https://github.com/cozy/cozy-ui/pull/1311). And in https://github.com/cozy/cozy-settings/pull/269/commits/8f8ce042ce206daca64194c0bfc726c615a7abca I used `Input` for cozy-ui, so I removed stylus code that added borders to all inputs (and required to manually reset the borders in some contexts)

package.json
drazik: Actually I think it's cozy-ui that requires cozy-doctypes, because the `ContactPicker` (not used here) needs it (see https://github.com/cozy/cozy-ui/blob/master/react/ContactPicker/index.jsx#L5). It's a peer dep, but when you `import { something } from 'cozy-ui'`, all imports are executed and webpack tries to find cozy-doctypes, and fails.

src/components/Input.jsx
ptbrowne: The matching of a regex can also be fairly expensive.

src/containers/Passphrase.jsx
drazik: Nothing more than "I copy/pasted it from another file and left it as is". It's totally possible to use await/async

src/containers/Passphrase.jsx
drazik: See https://github.com/cozy/cozy-settings/pull/269/commits/bf320b8bdb79bb421577a521f02e426730ede871

src/containers/Passphrase.jsx
drazik: See https://github.com/cozy/cozy-settings/pull/269/commits/819ef5ef41080821db2b288feda4b4599ea34e2d

src/containers/Passphrase.jsx
drazik: See https://github.com/cozy/cozy-settings/pull/269/commits/2f531491dd22f5685ff995fbe3d0a404b3c51633

src/containers/Passphrase.jsx
drazik: See https://github.com/cozy/cozy-settings/pull/269/commits/2f531491dd22f5685ff995fbe3d0a404b3c51633

PR #275, opened 2020-01-15
fix(ProfileView): Add missing translations

drazik:
Some translations were missing. This fixes it


Crash--:
Is it done on transifex ?

drazik:
Yep

PR #276, opened 2020-01-20
feat(devices): Show devices with "browser" kind

drazik:
This kind of clients includes the cozy browser extension

image


src/components/DevicesView.jsx
```diff
@@ -96,11 +104,15 @@ class DevicesView extends Component {
...
+                        tableStyles['coz-table-primary']
                       )}
                     >
-                      {device.client_name}
+                      <Media>
+                        <Img>
+                          <Icon icon={getDeviceIcon(device)} size={32} />
+                        </Img>
+                        <Bd className="u-ml-1">{device.client_name}</Bd>
+                      </Media>
```
Crash--: Should we use a CompositeRow instead?

drazik: This is really a table row. The use of Media here is to vertically align things without adding custom classes.

PR #288, opened 2020-07-03
feat: Use transpiled cozy-ui

ptbrowne:
See https://github.com/cozy/cozy-ui/issues/1492 for context


Crash--:
Settings already use the transpiled version of UI https://github.com/cozy/cozy-settings/blob/master/config/webpack.config.cozy-ui.js#L6

But personally I prefer to have the explicit import instead of relying on an alias.

PR #290, opened 2020-10-13
Updates and improvements

ptbrowne:
- Update dependencies
  - cozy-scripts
  - cozy-client
  - cozy-ui

- Remove cozy-client-js
  - Use cozy-interapp for intents ✅

- Fix small bugs leading to non working intents due to previous refactors
- Add test for sending email


Crash-- (approved):
CI is red, but looks solid! Thanks for this work!

src/actions/domUtils.js
```diff
@@ -0,0 +1,17 @@
...
+      '//' + document.querySelector('[role=application]').dataset.cozyDomain
+    )
+  } else {
+    return null
+  }
+}
+
+export const getStackToken = () => {
+  if (document.querySelector('[role=application]')) {
+    return '//' + document.querySelector('[role=application]').dataset.cozyToken
```
nono, line 13: I don't think we should have `'//' +` for the token

ptbrowne: Fixed

PR #317, opened 2020-10-28
Allow to import the data from another Cozy

nono:
The goal is to allow users to import the data exported from another Cozy instance. We need https://github.com/cozy/cozy-stack/pull/2688 to have a functional import on the stack.

The integration of the mockups is not faithful, we are waiting work by @JF-Cozy on `Dialog` for that.

#### Checklist

- [ ] Faithful integration of the mockups at all screen sizes
- [ ] Tested on supported browsers, including responsive mode (IE11 and Edge, Chrome >= 42, 2 latest versions of Firefox and Safari)
- [x] Localized in english and french


src/components/Import.jsx
```diff
@@ -0,0 +1,119 @@
...
+                {t('ProfileView.import.modal.title')}
+              </h2>
+            </ModalHeader>
+            <ModalDescription
+              className={viewStyles['set-export-modal-description']}
+            >
+              <ReactMarkdownWrapper
+                source={t('ProfileView.import.modal.description')}
+              />
+              <MuiCozyTheme>
```
Crash--, line 81: You should not add it there. It should be added at the top of the React Tree. I think https://github.com/cozy/cozy-settings/blob/master/src/targets/browser/index.jsx#L56 is a good place 👍

src/components/export/ExportSection.jsx
```diff
@@ -48,7 +48,7 @@ class ExportSection extends Component {
     const { displayModal } = this.state
     return (
       <div>
-        <div className={viewStyles['set-view-section']}>
+        <div className="{viewStyles['set-view-section']} u-mb-1">
```
Crash--, line 51:
```jsx 
import classNames from 'classnames'
```

```suggestion
        <div className={classNames(viewStyles['set-view-section'], 'u-mb-1')}>
```

nono: Why this change? It's more verbose but I don't understand what we get in exchange

Crash--: I read the diff from the github and didn't see the closing bracket, so I thought that it should not work. 

Beside the fact it's easier for me to read, we get nothing.

src/components/Import.jsx
ptbrowne: Please add it also for the intents, so that browser/index.jsx and intsn/index.jsx stay mostly the same.

JF-Cozy (approved):

PR #332, opened 2020-12-09
Use Dialog instead of Modal

ptbrowne:
Dialog from cozy-ui is simpler to use and has consistent behavior
on all our apps.

Also fixed some bugs along the way

- Loading text in Spinner was incorrectly displayed
- Compliqué was incorrectly displayed in 2FA modal


src/components/DevicesModaleRevokeView.jsx
```diff
@@ -1,12 +1,11 @@
 import React from 'react'
+import ReactMarkdown from 'react-markdown'
```
Crash--, line 2: Quoi ? T'as pas changé ça ? :trollface:

Crash-- (approved):
Thanks!

PR #345, opened 2021-03-16
Support for delegated authentication

ptbrowne:
- Hide password section when cozy only supports delegated authentication
- Modify password page title when cozy only supports delegated authentication
- Various small cosmetic improvements


src/targets/browser/index.ejs
```diff
@@ -31,6 +31,7 @@
   data-cozy-tracking="{{.Tracking}}"
   data-cozy-app-editor="{{.AppEditor}}"
   data-cozy-icon-path="{{.IconPath}}"
+  data-cozy="{{.CozyData}}"
```
Crash--, line 34: hmmm, since you don't modify the `new CozyClient` does that means that CC will check itself in the DOM to get the data?

Crash--: To myself, yes and no... : there is a loadFromDom() call https://github.com/cozy/cozy-client/blob/69e5a405c7c9d2b01b36ee6c389c48a2a51b063c/packages/cozy-client/src/CozyClient.js#L181 but there is no login if there is no option. So how does it work?

src/lib/client.js
```diff
@@ -1,6 +1,6 @@
 import CozyClient from 'cozy-client'
 
-const client = new CozyClient({
+const client = CozyClient.fromDOM({
```
Crash--, line 3: Aaaah, makes sense now! So previously we had an instance of CozyClient without having it 'logged' or else 🤔

src/components/ProfileView.jsx
```diff
@@ -24,21 +24,17 @@ const PasswordSection = () => {
...
+        {t('ProfileView.password.title')}
+      </Typography>
+      <Typography variant="body1">{t('ProfileView.password.label')}</Typography>
+      <Button
+        tag={Link}
+        to="/profile/password"
+        label={t('ProfileView.password.cta')}
+        theme="secondary"
+        className="u-mt-half u-mh-0"
+      />
```
Crash--, line 44: can you squash or rename this commit?

src/components/ProfileView.jsx
```diff
@@ -21,8 +24,9 @@ import { AUTH_MODE } from 'actions/twoFactor'
 
 const PasswordSection = () => {
   const { t } = useI18n()
-
-  return (
+  const client = useClient()
+  const canAuthWithPassword = get(client, 'capabilities.can_auth_with_password')
```
Crash--, line 31: capabalities.can_auth... is injected by the Stack in cozy-data, right? So I imagine that we do : `new CozyClient(..., { capabilities : { can_auth...`, but where do we get access to this property from client itself? Shouldn't be get(client.options, 'capabilities'...) ?

src/components/PassphraseForm.jsx
```diff
@@ -87,9 +91,18 @@ class PassphraseForm extends Component {
       location.search
     )
 
+    const canAuthWithPassword = get(
+      client,
+      'capabilities.can_auth_with_password'
+    )
+    const canAuthWithOIDC = get(client, 'capabilities.can_auth_with_oidc')
+    const isOIDC = !canAuthWithPassword && canAuthWithOIDC
```
Crash--, line 99: can't we have both? I mean having a canAuthWithOIDC but also having a password?

Crash-- (approved):

src/components/PassphraseForm.jsx
ptbrowne: I think we can have both, and if we have both we do not want to have the OIDC title.

src/components/ProfileView.jsx
ptbrowne: I had not yet opened this PR : https://github.com/cozy/cozy-client/pull/904 ;)

src/lib/client.js
ptbrowne: I don't understand the " without having it 'logged' or else 🤔" part ?

src/components/ProfileView.jsx
ptbrowne: Done

src/targets/browser/index.ejs
ptbrowne: Solved later on via CozyClient.fromDOM

src/components/PassphraseForm.jsx
Crash--: OK thanks for the explanation. So a small nit: `shouldUseOIDCtitle` : ... :D

PR #347, opened 2021-03-17
Fix: Capabilities

ptbrowne:
If the stack does not yet put capabilities in the server variable
CozyData, we need to have the "update password" button shown.

### DIFF
--- a/src/components/ProfileView.jsx
+++ b/src/components/ProfileView.jsx
@@ -1,6 +1,7 @@
 import React, { Component } from 'react'
 import { Link } from 'react-router-dom'
 import compose from 'lodash/flowRight'
+import get from 'lodash/get'
 
 import { useClient } from 'cozy-client'
 import { translate } from 'cozy-ui/transpiled/react/I18n'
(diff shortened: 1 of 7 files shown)


src/components/ProfileView.jsx
```diff
@@ -28,7 +26,10 @@ import { AUTH_MODE } from 'actions/twoFactor'
 const PasswordSection = () => {
   const { t } = useI18n()
   const client = useClient()
-  const canAuthWithPassword = get(client, 'capabilities.can_auth_with_password')
+  const capabilities = client.capabilities
+  const canAuthWithPassword = capabilities
+    ? capabilities.can_auth_with_password
+    : true // If we have no capabilities, consider that we can login to the cozy with password
```
Crash--, line 32: you can also do 
```
const canAuthWithPassword = get(client, 'capabilities.can_auth_with_password', true /* If we have no capabilities, consider that we can login to the cozy with password*/)
```

Crash-- (approved):

src/components/ProfileView.jsx
ptbrowne: Thanks, I thought about that but was not sure if extra values would not be considered as extra keys. I've made the change 👍

Crash-- (approved):

PR #356, opened 2021-06-28
fix: Fix password strength computation

nono:
The password strength indicator uses the classes of the characters in
the password, but some characters weren't classified correctly:

- `É` is an uppercase letter
- `$` is a symbol


Crash-- (approved):
Just a linter issue but :+1:

PR #357, opened 2021-09-13
WIP

taratatach:
*Please explain what this PR does here.*

#### Checklist

- [ ] Faithful integration of the mockups at all screen sizes
- [ ] Tested on supported browsers, including responsive mode (IE11 and Edge, Chrome >= 42, 2 latest versions of Firefox and Safari)
- [ ] Localized in english and french

If this PR relies on changes in one of our internal libraries (cozy-client-js or cozy-ui, for example), please also do the following things:

- [ ] The package.json and yarn.lock files include the correct dependency version, in those versions have been released
- [ ] [cozy-ui](https://github.com/cozy/cozy-ui) documentation and style guides are up to date

*(If not, please remove this last section.)*


src/components/DevicesModaleConfigureView.jsx
```diff
@@ -0,0 +1,305 @@
...
+  toggleInclusion,
+  toggleAllInclusion
+}) => {
+  const { t } = useI18n()
+
+  return (
+    <Card className="u-mt-1 u-p-0">
+      <CardHeader
+        title={
+          <Typography variant="subtitle1" className="u-coolGrey">
```
Crash--, line 96: We avoid the use of non semantic color. I think 
```suggestion
          <Typography variant="subtitle1" color="textSecondary">
```

should do the work https://docs.cozy.io/cozy-ui/react/#!/Typography

src/components/DevicesModaleConfigureView.jsx
```diff
@@ -0,0 +1,305 @@
...
+            mixed={excluded.length > 0 && excluded.length !== folders.length}
+            disabled={!partialSync}
+            onClick={() => {
+              setPartialSync(true)
+              toggleAllInclusion()
+            }}
+          />
+        }
+        classes={{
+          root: 'u-pv-half u-ph-1-half u-bg-paleGrey u-fz-tiny',
```
Crash--, line 112:
```suggestion
          root: 'u-pv-half u-ph-1-half u-bg-paleGrey u-fz-tiny',
```

Same for paleGrey. You can talk with @JF-Cozy or @joel-costa to know the var to use.

src/components/DevicesModaleConfigureView.jsx
```diff
@@ -0,0 +1,305 @@
...
+  const foldersQuery = buildFoldersQuery({ currentFolderId: ROOT_FOLDER_ID })
+  const queryResult = useQuery(foldersQuery.definition, foldersQuery.options)
+
+  useMemo(() => {
+    // We wait until the query has been reloaded (and thus data has been fetched
+    // again) before marking our componant as loaded to avoid glitches between
+    // stale data loaded from the store and new data fetched from the Cozy.
+    if (!wasLoading && isQueryLoading(queryResult)) {
+      setWasLoading(true)
+    } else if (wasLoading && !loaded && !isQueryLoading(queryResult)) {
```
Crash--, line 157: if the request fails the first time, we'll pass in that block, right? I think this can be avoided by using `hasQueryBeenLoaded` instead?

src/components/DevicesModaleConfigureView.jsx
```diff
@@ -0,0 +1,305 @@
...
+      const wasExcluded =
+        queryResult.data && queryResult.data.length > 0
+          ? queryResult.data.filter(
+              folder =>
+                folder.not_synchronized_on != null &&
```
Crash--, line 162:
```suggestion
                folder.not_synchronized_on !== null &&
```

?

src/components/DevicesModaleConfigureView.jsx
```diff
@@ -0,0 +1,305 @@
...
+                folder.not_synchronized_on.find(({ id }) => id === device.id) !=
```
Crash--, line 163:
```suggestion
                folder.not_synchronized_on.find(({ id }) => id === device.id) !==
```

src/components/DevicesModaleConfigureView.jsx
taratatach: The single `=` is on purpose, otherwise tis wouldn't return `true` if `folder.not_synchronized_on` is `undefined`.

src/components/DevicesModaleConfigureView.jsx
taratatach: Same idea here, the single `=` is on purpose.

src/components/DevicesModaleConfigureView.jsx
taratatach: Thanks for the tip. I missed this prop.

src/components/DevicesModaleConfigureView.jsx
taratatach: `hasQueryBeenLoaded` does something entirely different as it only checks if the query result has a `lastFetch` attribute value.
But I can re-order the conditions because you're right, we'd go through this block if the query failed.

taratatach: (Done. The failed query conditional block now comes before this one.)

src/components/DevicesModaleConfigureView.jsx
Ldoppea: Another way to write `folder.not_synchronized_on != null` is `!!folder.not_synchronized_on`. This would also return `false` if `0` but my understanding here is that you are waiting and array, so it's fine

src/components/DevicesModaleConfigureView.jsx
Ldoppea: Here you can replace `find()` by `some()`. This will directly return a boolean if something matches the condition and then you wouldn't have to compare with `null`.

src/locales/fr.json
```diff
@@ -267,6 +268,19 @@
     "cancel": "Annuler",
     "error": "Je ne parviens pas à révoquer l’appareil, essayez de recharger la page."
   },
+  "configureDevice": {
+    "load_error": "Une erreur s’est produite pendant le chargement de vos dossiers, merci de réessayer prochainement.",
+    "configure_error": "Je ne parviens pas à configurer les dossiers à synchroniser avec votre ordinateur, essayez de recharger la page.",
+    "description": "Sélectionner les dossiers à synchroniser avec votre ordinateur :",
+    "no_folders": "Vous n'avez aucun dossier dans votre Cozy.",
+    "my_cozy": "Mon Cozy",
```
Ldoppea, line 276: :seedling: I should definitively check if unbreakable spaces are used in cozy-pass branding

src/components/DevicesModaleConfigureView.jsx
```diff
@@ -0,0 +1,338 @@
...
+)
+
+const FolderCheckbox = ({ onClick, checked, mixed, disabled, ...props }) => (
+  <IconButton size="small" onClick={onClick}>
+    <Checkbox
+      checked={checked}
+      mixed={mixed}
+      disabled={disabled}
+      disableRipple={true}
+      readOnly
```
Ldoppea, line 52: :question: What is the goal of `readOnly` here?

src/components/DevicesModaleConfigureView.jsx
```diff
@@ -0,0 +1,338 @@
...
+        }
+        action={
+          hasFolders && (
+            <FolderCheckbox
+              data-testid="toggle-all-inclusion"
+              checked={excluded.length !== folders.length}
+              mixed={excluded.length > 0 && excluded.length !== folders.length}
+              disabled={!partialSync}
+              onClick={() => {
+                setPartialSync(true)
```
Ldoppea, line 121: :thinking: This call seems unnecessary as `toggleAllInclusion()` will recompute this value

src/locales/fr.json
```diff
@@ -267,6 +268,19 @@
     "cancel": "Annuler",
     "error": "Je ne parviens pas à révoquer l’appareil, essayez de recharger la page."
   },
+  "configureDevice": {
+    "load_error": "Une erreur s’est produite pendant le chargement de vos dossiers, merci de réessayer prochainement.",
+    "configure_error": "Je ne parviens pas à configurer les dossiers à synchroniser avec votre ordinateur, essayez de recharger la page.",
```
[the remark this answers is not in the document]
taratatach: It's out of the scope of this PR but feel free to bring it up on our channel or to create a ticket :)

src/components/DevicesModaleConfigureView.jsx
taratatach: Good catch! I forgot to remove those after I introduced `updateExcluded`.

src/components/DevicesModaleConfigureView.jsx
taratatach: If I remember correctly, I was hoping this would remove the halo around the checkbox. I found it in `cozy-ui` on some checkbox.
It's not documented anywhere so it's probably not doing anything and it apparently does nothing on the halo.

src/lib/withFolders.jsx
```diff
@@ -0,0 +1,50 @@
...
+  const [folders, setFolders] = useState([])
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        setLoading(true)
+        const foldersQuery = buildFoldersQuery({
+          currentFolderId: ROOT_FOLDER_ID
+        })
+        const fetchedFolders = await client.queryAll(foldersQuery)
```
Crash--, line 36: We don't like the use of queryAll(). It can be very slow even if I don't think we'll have a lot of cases with thousands and thousands of folders. 

The way we handle that is to make the first `query`. Then,  the query returns an `hasMore` attribute. If this value if `true` then we should call `fetchMore()` only if needed. And when I say that, it means, we create an Intersection Observer https://developer.mozilla.org/fr/docs/Web/API/Intersection_Observer_API

We have a Load More button that do that for you: https://github.com/cozy/cozy-ui/blob/master/react/LoadMore/index.jsx

And you can see its usage in Drive here: https://github.com/cozy/cozy-drive/blob/master/src/drive/web/modules/views/Folder/FolderViewBody.jsx#L212-L214

src/components/DevicesModaleConfigureView.jsx
```diff
@@ -0,0 +1,306 @@
...
+  return (
+    <Card className="u-mt-1 u-p-0">
+      <CardHeader
+        title={
+          <Typography variant="subtitle1" color="textSecondary">
+            {t('configureDevice.my_cozy')}
+          </Typography>
+        }
+        action={
+          hasFolders && (
```
Merkur39, line 106: Je pense qu'il est préférable de renvoyer `null` dans le cas où `hasFolders == false`
Du genre:
```
hasFolders ? (
  <FolderCheckbox ... />
) : null
```

taratatach: Hum, quelle différence avec le fait de renvoyer `undefined` ?

src/lib/flags.js
```diff
@@ -0,0 +1,22 @@
+import flag from 'cozy-flags'
+
+export const initFlags = () => {
+  let activateFlags = flag('switcher') === true ? true : false
+
+  if (process.env.NODE_ENV !== 'production' && flag('switcher') === null) {
```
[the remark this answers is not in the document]
taratatach: Ce n'est pas équivalent. Par exemple, si `flag('switcher')` renvoie `false`, alors `!!!flag('switcher')` renverra `true` alors que `flag('switcher') === null` renverra `false`.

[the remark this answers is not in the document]
Ldoppea: There seems to be too many `!` on your example. Maybe just `!!flag('switcher')` ?

Also `let activateFlags = flag('switcher') === true` is valid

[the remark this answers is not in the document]
Ldoppea: Woups je répondais au premier commentaire lié au `=== true`, pas au `=== null`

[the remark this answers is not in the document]
Ldoppea: Donc le commentaire sur le !!! n'est pas d'actu

src/components/DevicesModaleConfigureView.jsx
Ldoppea: Ca dépend de l'implémentation du composant CardHeader, s'il tente d'afficher `action` sans vérification il va tenter de rendre `undefined`. Hors React ne sais pas rendre `undefined`, par contre il sais rendre `null` : https://fr.reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering

Mais dans ce cas pas de problème car MaterialUI vérifie qu'il est bien défini avant de l'utiliser:
https://github.com/mui-org/material-ui/blob/master/packages/mui-material/src/CardHeader/CardHeader.js#L141

src/lib/flags.js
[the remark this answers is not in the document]
Merkur39: En effet, mes excuses 😇 
Du coup, plus quelque chose du genre `!!flag('switcher') === false`
L'idée est de moins jongler entre les `falsy` (mais bon ça reste du "nit" hein 😁 )

src/components/DevicesModaleConfigureView.jsx
Merkur39: En effet, mais pour ma part je conseillerai de faire ce check au plus tôt, et ça évite de bétonner de mauvaises pratiques 😁

src/lib/flags.js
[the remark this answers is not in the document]
taratatach: Pas super fan des `!!` assez difficiles à lire au final mais ta proposition pour `activateFlags` me semble pertinente.

src/components/DevicesModaleConfigureView.jsx
taratatach: Hum OK. Pourtant ça fonctionne et je n'ai jamais eu de problème avec cette syntaxe :thinking:. C'est même proposé sur la même page de doc que tu donnes : https://fr.reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator

Merkur39: Oui dans ton exemple, tout comme dans ton cas finalement, ce n'est pas impactant car ce n'est pas le plus haut parent 👍

Ldoppea: Le cas qui pourrait planter c'est si de leur côté ils avaient fait un composant du style
```jsx
function CardHeader({ action }) {
  return action && <div>lorem ipsum</div>
}
```

Le fait de retourner `undefined` aurait posé problème mais `null` aurait été accepté.

Par ceci fonctionne (car on retourne quelque chose)
```jsx
function CardHeader({ action }) {
  return <div>{action}</div>
}
```

package.json
```diff
@@ -15,7 +15,7 @@
     "start": "cozy-scripts start --hot",
     "deploy": "cs deploy",
     "cozyPublish": "cozy-scripts publish --token $REGISTRY_TOKEN --prepublish downcloud --postpublish mattermost",
-    "test": "cozy-scripts test",
+    "test": "cozy-scripts test --env=jest-environment-jsdom-sixteen",
```
Ldoppea, line 18: Note that this wouldn't be needed with a newer version of jest in create-cozy-app. Would it be a thing to upgrade it in our projects?

https://github.com/testing-library/dom-testing-library/releases/tag/v7.0.0

src/components/DevicesModaleConfigureView.spec.jsx
```diff
@@ -0,0 +1,771 @@
...
+      })
+    })
+  })
+
+  describe('when the folders query failed', () => {
+    beforeEach(() => {
+      useFolders.mockReturnValue({
+        loading: false,
+        failed: true,
+        folders: mockFolders // XXX: just checking that this has no effects on the UI
```
Ldoppea, line 117: What's the meaning of XXX?

Ldoppea (approved):

src/components/DevicesModaleConfigureView.spec.jsx
taratatach: It's a common pattern to catch the eye (my editor highlights it for example). You can see others are using it https://stackoverflow.com/questions/1452934/what-is-the-meaning-of-xxx-in-code-comments but we use it at least in the stack and desktop code bases.
I use it as a note when explaining something that could seem rather odd.

package.json
Crash--: https://github.com/cozy/create-cozy-app/issues/1419

src/components/DevicesModaleConfigureView.jsx
taratatach: OK, je comprends mieux. Je fais ce changement et je merge.

PR #379, opened 2021-12-16
feat: Configure Desktop from version 3.32.0-beta.2

taratatach:
The partial synchronization configuration should only be available for
Desktop clients with a minimum version of 3.32.0-beta.2 as earlier
versions require a flag to be activated to handle partial
synchronization or don't handle it at all.

We still require the flag to be activated for the button to be shown
though.

We add a new dependency to `semver` to help us compare the Desktop 
clients' version with our minimum version.

#### Checklist

- [x] Faithful integration of the mockups at all screen sizes
- [x] Tested on supported browsers, including responsive mode (IE11 and Edge, Chrome >= 42, 2 latest versions of Firefox and Safari)
- [ ] Localized in english and french

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
     "redux": "4.0.1",
     "redux-logger": "3.0.6",
     "redux-thunk": "2.3.0",
+    "semver": "^7.3.5",
     "ua-parser-js": "0.7.19"
   }
 }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -64,6 +64,7 @@
     "redux": "4.0.1",
     "redux-logger": "3.0.6",
     "redux-thunk": "2.3.0",
+    "semver": "^7.3.5",
```
Crash--, line 67: What is the size added by this lib to the bundle? 

There is no bundlemoon on this repo yet

Crash-- (approved):

package.json
taratatach: The bundle went from 3552 Bytes to 3580 :)

PR #385, opened 2022-02-11
fix(deps): remove unused date-fns

trollepierre:
date-fns is not used in the app (it is still present in yarn.lock because used in dependencies)

close #261


Crash-- (approved):
\o/

PR #397, opened 2022-06-24
fix: Update Mattermost channel

Noemie21:
Notifications of repository publications were sent in an old channel that is no longer used


Crash--:
When we'll merge this PR, can you check if the message is really posted on the right channel? I don't know if the mattermost token has the right to post in this new channel

Crash-- (approved):

PR #402, opened 2022-06-27
fix: Add flag for permissions tab

Noemie21:
Put the permissions tab behind a flag


paultranvan (approved):

Crash-- (changes requested):
Can you add the flag here please? https://github.com/cozy/cozy-settings/blob/44a84b61b2696645593c54d7bf056e1d88e51ef1/src/lib/flags.js#L20

trollepierre (approved):

src/lib/flags.js
```diff
@@ -19,5 +19,6 @@ export const initFlags = () => {
 
 const flagsList = () => {
   flag('switcher', true)
-  flag('settings.partial-desktop-sync.show-synced-folders-selection')
+  flag('settings.partial-desktop-sync.show-synced-folders-selection'),
+    flag('settings.permissions-dashboard', true)
```
trollepierre, line 23: are you sure, this is working?

according to the code syntax, I'm pretty sure, it does not.

do you know how to see this flaglist? if not, let's have a look together 🙏

Crash--: > according to the code syntax, I'm pretty sure, it does not.

According to the code, I'm pretty sure it works: 
https://github.com/cozy/cozy-libs/blob/master/packages/cozy-flags/src/flag.js#L9-L17

This flaglist is just here to be called during the init phase of the app. Like that, flags are set. And flags are now displayed in the FlagSwitcher https://github.com/cozy/cozy-libs/tree/master/packages/cozy-flags#demo and now you can play with the flag by clinking on stuff instead of playing with the console.

trollepierre: Apparently, not:
Capture d’écran 2022-06-29 à 12 34 15

I thought it was because of the comma, but removing it, does not change the flag seen in the flag debugger. 🤔

PR #406, opened 2022-07-06
feat: Password hint from required to optional

vdnj:
Change placeholder
Button is now clickable even if there is no hint

### DIFF
--- a/src/components/PassphraseForm.jsx
+++ b/src/components/PassphraseForm.jsx
@@ -82,7 +82,6 @@ class PassphraseForm extends Component {
       newPassphraseTouched &&
       newPassphraseMatch &&
       strength.label !== 'weak' &&
-      hint &&
       !hintSameAsPassphrase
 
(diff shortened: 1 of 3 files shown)


src/locales/en.json
```diff
@@ -315,7 +315,7 @@
     },
     "hint": {
       "title": "Leave a hint",
-      "placeholder": "Hint",
+      "placeholder": "Hint (optional)",
```
trollepierre, line 318: should we keep this optional?

src/components/PassphraseForm.jsx
```diff
@@ -82,7 +82,6 @@ class PassphraseForm extends Component {
       newPassphraseTouched &&
       newPassphraseMatch &&
       strength.label !== 'weak' &&
-      hint &&
```
trollepierre, line 85: the commit prefix that respects the convention is lowercase.

feat: Password...

trollepierre: btw "optional" with one "n"

PR #410, opened 2022-07-11
feat: Display app page on click

vdnj:
On permission page, if an app is clicked we are redirected to this app page.
With @Noemie21


src/containers/PermissionsApplication.jsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react'
```
trollepierre, line 1: ajoutons une modif dans le changelog 🙏

trollepierre (approved):

PR #411, opened 2022-07-13
feat(permissions): Display permissions on PermissionsApplication page

Noemie21:
with @vdnj 

Capture d’écran 2022-07-13 à 13 35 11

### DIFF
--- a/src/containers/PermissionsApplication.jsx
+++ b/src/containers/PermissionsApplication.jsx
@@ -29,7 +29,18 @@ const PermissionsApplication = ({ match }) => {
           {t('Permissions.failedRequest')}
         </Typography>
       ) : (
-        <PageTitle>{appName}</PageTitle>
+        <div>
+          <PageTitle>{appName.toUpperCase()}</PageTitle>
+          {Object.entries(queryResult.data[0].attributes.permissions).map(
+            ([key, value]) => (
+              <React.Fragment key={key}>
+                <Typography variant="h4">
+                  {value.type} : {value.verbs ? value.verbs.join(' / ') : 'ALL'}
+                </Typography>
+              </React.Fragment>
+            )
+          )}
+        </div>
       )}
     </Page>
   )
--- a/src/containers/PermissionsApplication.spec.jsx
+++ b/src/containers/PermissionsApplication.spec.jsx
@@ -39,10 +39,35 @@ jest.mock('cozy-ui/transpiled/react/Spinner', () => {
 
 describe('PermissionsApplication', () => {
   beforeEach(() => {
+    const queryResult = {
+      fetchStatus: 'loaded',
+      data: [
+        {
+          attributes: {
+            permissions: {
+              files: {
+                type: 'io.cozy.files',
+                description: 'Required to access the files'
+              },
+              allFiles: {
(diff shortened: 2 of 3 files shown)


src/containers/PermissionsApplication.jsx
```diff
@@ -29,7 +29,21 @@ const PermissionsApplication = ({ match }) => {
...
         </Typography>
       ) : (
-        <PageTitle>{appName}</PageTitle>
+        <div>
+          <PageTitle>{appName.toUpperCase()}</PageTitle>
+          {Object.entries(queryResult.data[0].attributes.permissions).map(
+            ([key, value]) => (
+              <React.Fragment key={key}>
+                <Typography variant="h4">{key}</Typography>
+                {value.verbs && (
```
trollepierre, line 38: can we add inside the snapshot, an App, with a value containing verbs please?

src/containers/PermissionsApplication.jsx
```diff
@@ -29,7 +29,21 @@ const PermissionsApplication = ({ match }) => {
           {t('Permissions.failedRequest')}
         </Typography>
       ) : (
-        <PageTitle>{appName}</PageTitle>
+        <div>
+          <PageTitle>{appName.toUpperCase()}</PageTitle>
+          {Object.entries(queryResult.data[0].attributes.permissions).map(
+            ([key, value]) => (
```
trollepierre, line 35: i think the commit should be named like the name of the trello story. We are not only displaying permissions, but also permissions details, inside PermApplication page.

src/containers/PermissionsApplication.jsx
```diff
@@ -29,7 +29,21 @@ const PermissionsApplication = ({ match }) => {
           {t('Permissions.failedRequest')}
         </Typography>
       ) : (
-        <PageTitle>{appName}</PageTitle>
+        <div>
+          <PageTitle>{appName.toUpperCase()}</PageTitle>
```
Crash--, line 33: nit: This is not really the appName. This is the `appSlug`. 

Exemple: https://github.com/cozy/cozy-drive/blob/master/src/drive/targets/manifest.webapp#L2-L4

If you really want, you can display the real appName: https://github.com/cozy/cozy-drive/blob/master/src/drive/targets/manifest.webapp#L2-L4 https://github.com/cozy/cozy-notes/blob/master/manifest.webapp#L2-L4 . You'll not need this toUpperCase() method.

src/containers/PermissionsApplication.jsx
Crash--: @Noemie21 : Sometimes you'll not have any verbs. When not having verbs it means that you have all the rights. This is the same as "ALL" or the same as "GET / PUT / POST / DELETE"  

Ex: https://github.com/cozy/cozy-notes/blob/master/manifest.webapp#L43-L46

Crash--: Also, I think that the important stuff is to displayed the concerned doctype (aka the type) instead of the `key`.

I would prefer to see: 
io.cozy.files: ALL 
io.cozy.contacts: GET 

instead of: 
allFiles: ALL
Contact: GET

PR #426, opened 2022-08-01
feat: Add number of permissions

Noemie21:


src/containers/Permissions.jsx
```diff
@@ -50,19 +50,29 @@ const Permissions = () => {
...
+                      <ListItemIcon>
+                        <AppIcon app={appOrKonnector} />
+                      </ListItemIcon>
+                      <ListItemText
+                        primary={appOrKonnector.name}
+                        secondary={
+                          appOrKonnector.permissions
+                            ? Object.entries(appOrKonnector.permissions)
+                                .length + ' permissions'
+                            : '0' + ' permissions'
```
trollepierre, line 67: it should be possible to handle the plural with transifex. (when there is one permission)

src/containers/Permissions.spec.jsx
```diff
@@ -117,11 +130,19 @@ describe('Permissions', () => {
...
+    expect(container).toMatchSnapshot()
+    expect(getAllByTestId('ListItemText')[1]).toHaveAttribute(
+      'data-primary',
+      'Contacts'
+    )
+    expect(getAllByTestId('ListItemText')[0]).toHaveAttribute(
+      'data-primary',
+      'Alan'
+    )
+    // expect(queryByText('Alan')).toBeTruthy()
```
trollepierre, line 145: remove this line 🙏

src/containers/__snapshots__/Permissions.spec.jsx.snap
```diff
@@ -32,14 +32,41 @@ exports[`Permissions should display appName when query has been loaded 1`] = `
               />
             </div>
             <div
-              class="MuiListItemText-root"
+              data-primary="Alan"
+              data-secondary="0 permissions"
```
trollepierre, line 36: would that be possible to check in one test, when there are permissions? (instead of  0 permissions)

src/containers/Permissions.jsx
trollepierre: Referring to https://github.com/cozy/cozy-drive/blob/master/src/drive/locales/en.json#L259

Crash--: We're using the node version of polyglot https://airbnb.io/polyglot.js/ but this is the same thing. You can use a `%{smart_count}` variable ;)

src/containers/Permissions.spec.jsx
paultranvan: @Noemie21 resolved by still there?

src/containers/Permissions.spec.jsx
```diff
@@ -76,23 +85,7 @@ describe('Permissions', () => {
...
-                description: 'Required to access the files'
-              },
-              konnectors: {
-                description: 'Required to get the list of konnectors',
-                type: 'io.cozy.konnectors',
-                verbs: ['GET']
-              }
-            }
-          },
+          permissions: {},
```
trollepierre, line 88: is it possible, a konnector with no permission? 🤔

src/locales/en.json
```diff
@@ -398,6 +398,8 @@
   
   "Permissions": {
     "title": "Applications control",
-    "failedRequest": "The request has failed"
+    "failedRequest": "The request has failed",
+    "numberOfPermissions" : "%{smart_count} permission |||| %{smart_count} permissions",
+    "noPermissions": "0 permission"
```
trollepierre, line 403: why no Permission is used  like numberOfPermissions ?

trollepierre (approved):

src/containers/Permissions.spec.jsx
Crash--: It'll do nothing, but there is nothing preventing it.

yarn.lock
```diff
@@ -3768,6 +3768,11 @@ aria-query@^5.0.0:
   resolved "https://registry.yarnpkg.com/aria-query/-/aria-query-5.0.0.tgz#210c21aaf469613ee8c9a62c7f86525e058db52c"
   integrity sha512-V+SM7AbUwJ+EBnB8+DXs0hPZHO0W6pqBcc0dW90OwtVG02PswOu/teuARoLQjdDOH+t9pJgGnW5/Qmouf3gPJg==
 
+aria-query@^5.0.0:
+  version "5.0.0"
```
trollepierre, line 3772: why this?

PR #428, opened 2022-08-03
fix: Remove Home, Store and Settings from the list of apps

Noemie21:
When the user uses his Cozy, he does not see the home, settings or the store as applications but as the Cozy platform. Technically they are definitely apps but they are not seen as such functionally. That's why they should not be displayed

### DIFF
--- a/src/containers/Permissions.jsx
+++ b/src/containers/Permissions.jsx
@@ -31,6 +31,7 @@ const Permissions = () => {
     as: KONNECTORS_DOCTYPE,
     fetchPolicy: CozyClient.fetchPolicies.olderThan(THIRTY_SECONDS)
   })
+  const toNotDisplay = ['home', 'store', 'settings']
 
   return (
     <Page narrow>
@@ -50,6 +51,7 @@ const Permissions = () => {
           {queryResultApps.data
             .concat(queryResultKonnectors.data)
             .sort((a, b) => a.slug.localeCompare(b.slug))
+            .filter(a => !toNotDisplay.includes(a.slug))
             .map(appOrKonnector => {
               return (
                 <div key={appOrKonnector.name}>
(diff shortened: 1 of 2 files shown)


src/containers/Permissions.jsx
```diff
@@ -50,19 +51,22 @@ const Permissions = () => {
...
-                    <ListItemIcon>
-                      <AppIcon app={appOrKonnector} />
-                    </ListItemIcon>
-                    <ListItemText primary={appOrKonnector.name} />
-                  </ListItem>
-                </Link>
-                <Divider />
-              </div>
-            ))}
+            .map(
```
trollepierre, line 54: let's use `.filter()`

src/containers/Permissions.jsx
```diff
@@ -31,6 +31,7 @@ const Permissions = () => {
     as: KONNECTORS_DOCTYPE,
     fetchPolicy: CozyClient.fetchPolicies.olderThan(THIRTY_SECONDS)
   })
+  const toNotDisplay = ['Home', 'Store', 'Settings']
```
trollepierre, line 34: let's add those 3 applications inside queryAppsResult inside the tests, to check they are not present inside the DOM

trollepierre:
could you provide a description inside the commit, and in the PR description as well, explaining why we do this change 🙏 ?

src/containers/Permissions.jsx
```diff
@@ -50,19 +51,22 @@ const Permissions = () => {
...
-                    </ListItemIcon>
-                    <ListItemText primary={appOrKonnector.name} />
-                  </ListItem>
-                </Link>
-                <Divider />
-              </div>
-            ))}
+            .map(
+              appOrKonnector =>
+                !toNotDisplay.includes(appOrKonnector.name) && (
```
Crash--, line 56: You should not rely on the `name`. `name` is something that can change and can be overwritten depending on the Cozy's context. Use the `slug` attribute instead.

trollepierre (approved):

PR #436, opened 2022-08-08
feat: Add backward arrow in Permission

Noemie21:
Capture d’écran 2022-08-08 à 09 47 44

### DIFF
--- a/src/containers/__snapshots__/Permission.spec.jsx.snap
+++ b/src/containers/__snapshots__/Permission.spec.jsx.snap
@@ -7,6 +7,31 @@ exports[`Permission should display appName when query has been loaded 1`] = `
     data-testid="page"
   >
     <div>
+      <a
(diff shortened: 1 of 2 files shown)


src/containers/__snapshots__/Permission.spec.jsx.snap
```diff
@@ -7,6 +7,31 @@ exports[`Permission should display appName when query has been loaded 1`] = `
     data-testid="page"
   >
     <div>
+      <a
+        aria-disabled="false"
+        class="MuiButtonBase-root MuiIconButton-root u-mr-half large"
+        href="#/permissions/Drive"
```
trollepierre, line 13:
```suggestion
        href="#/permissions/drive"
```

=> I think we should use a lowercase in the appName

trollepierre (approved):

PR #438, opened 2022-08-09
feat: Add tab to display DataList

Noemie21:
The list of applications is now on the path '#/permissions/slug',
The data page is on the path '#/permissions/data'

Capture d’écran 2022-08-09 à 12 23 58

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -18,7 +18,7 @@ import Devices from 'containers/Devices'
 import Sessions from 'containers/Sessions'
 import Storage from 'containers/Storage'
 import Passphrase from 'containers/Passphrase'
-import Permissions from 'containers/Permissions'
+import PermissionsTab from './Permissions/PermissionsTab'
 import IntentRedirect from 'services/IntentRedirect'
 import PermissionsApplication from 'containers/PermissionsApplication'
 import Permission from 'containers/Permission'
@@ -44,18 +44,19 @@ export class App extends Component {
             <Route path="/connectedDevices/:deviceId" component={Devices} />
             <Route path="/sessions" component={Sessions} />
             <Route path="/storage" component={Storage} />
-            <Route exact path="/permissions" component={Permissions} />
+            <Route exact path="/permissions/:page" component={PermissionsTab} />
             <Route
               exact
-              path="/permissions/:app"
+              path="/permissions/slug/:app"
               component={PermissionsApplication}
             />
             <Route
-              path="/permissions/:app/:permission"
+              path="/permissions/slug/:app/:permission"
               component={Permission}
             />
             <Route path="/exports/:exportId" component={Profile} />
             <Redirect exact from="/" to="/profile" />
+            <Redirect exact from="/permissions" to="/permissions/slug" />
             <Redirect from="*" to="/profile" />
           </Switch>
         </Main>
--- a/src/components/Permissions/DataList.jsx
+++ b/src/components/Permissions/DataList.jsx
@@ -1,0 +1,15 @@
+import React from 'react'
+import Page from 'components/Page'
+import PageTitle from 'components/PageTitle'
+import { useI18n } from 'cozy-ui/transpiled/react/I18n'
+
+const DataList = () => {
+  const { t } = useI18n()
+  return (
+    <Page narrow>
+      <PageTitle>{t('Permissions.data')}</PageTitle>
+    </Page>
+  )
+}
+
+export default DataList
--- a/src/components/Permissions/DataList.spec.jsx
+++ b/src/components/Permissions/DataList.spec.jsx
@@ -1,0 +1,34 @@
+import { render } from '@testing-library/react'
+import React from 'react'
+import DataList from './DataList'
+import AppLike from 'test/AppLike'
+
+jest.mock('cozy-ui/transpiled/react', () => {
+  return { useI18n: () => ({ t: x => x }) }
+})
+
+jest.mock('components/Page', () => {
+  // eslint-disable-next-line react/display-name
+  return ({ narrow, children }) => (
+    <div data-testid="page" data-narrow={narrow}>
+      {children}
+    </div>
+  )
+})
+
+jest.mock('components/PageTitle', () => {
+  // eslint-disable-next-line react/display-name
+  return ({ children }) => <div data-testid="page-title">{children}</div>
+})
+
+describe('DataList', () => {
+  it('should display DataList title', () => {
+    const { queryByText, container } = render(
+      <AppLike>
+        <DataList />
+      </AppLike>
+    )
+    expect(container).toMatchSnapshot()
+    expect(queryByText('Data')).toBeTruthy()
+  })
+})
--- a/src/components/Permissions/PermissionsTab.jsx
+++ b/src/components/Permissions/PermissionsTab.jsx
@@ -1,0 +1,63 @@
+import React, { useState } from 'react'
+import Page from 'components/Page'
+import { Tabs, Tab } from 'cozy-ui/transpiled/react/MuiTabs'
+import AppList from './AppList'
+import DataList from './DataList'
+import Divider from 'cozy-ui/transpiled/react/MuiCozyTheme/Divider/index'
+import { routes } from 'constants/routes'
+import { withRouter } from 'react-router-dom'
+import withAllLocales from '../../lib/withAllLocales'
+
+function a11yProps(index) {
+  return {
+    id: `simple-tab-${index}`,
+    'aria-controls': `simple-tabpanel-${index}`
+  }
+}
+const TabPanel = props => {
+  const { children, value, index, ...other } = props
+  return (
+    <div
+      role="tabpanel"
+      hidden={value !== index}
+      id={`simple-tabpanel-${index}`}
+      aria-labelledby={`simple-tab-${index}`}
+      {...other}
+    >
+      {value === index && children}
+    </div>
+  )
+}
+const PermissionsTab = ({ match, t }) => {
+  const slugOrData = match.params.page
+  const [tab, setTab] = useState(slugOrData)
+  const handleChange = (event, value) => setTab(value)
+
+  return (
+    <Page narrow>
+      <Tabs value={tab} onChange={handleChange}>
+        <Tab
+          value="slug"
+          label={t('Permissions.applications')}
+          href={`#${routes.appList}`}
+          {...a11yProps(0)}
+        />
+        <Tab
+          value="data"
+          label={t('Permissions.data')}
+          href={`#${routes.dataList}`}
+          {...a11yProps(1)}
+        />
+      </Tabs>
+      <Divider className="u-mb-1" />
+      <TabPanel value={tab} index="slug">
+        <AppList />
+      </TabPanel>
+      <TabPanel value={tab} index="data">
+        <DataList />
+      </TabPanel>
+    </Page>
+  )
+}
+
+export default withRouter(withAllLocales(PermissionsTab))
--- a/src/components/Permissions/__snapshots__/AppList.spec.jsx.snap
+++ b/src/components/Permissions/__snapshots__/AppList.spec.jsx.snap
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Permissions should display appName when query has been loaded 1`] = `
+exports[`AppList should display appName when query has been loaded 1`] = `
 <div>
   <div
     data-narrow="true"
@@ -9,7 +9,7 @@ exports[`Permissions should display appName when query has been loaded 1`] = `
     <div
       data-testid="page-title"
     >
-      Permissions.title
+      Applications
     </div>
     <ul
       class="MuiList-root MuiList-padding"
@@ -33,7 +33,7 @@ exports[`Permissions should display appName when query has been loaded 1`] = `
             </div>
             <div
               data-primary="Alan"
-              data-secondary="Permissions.numberOfPermissions"
+              data-secondary="0 permissions"
               data-testid="ListItemText"
             />
             <span
@@ -64,7 +64,7 @@ exports[`Permissions should display appName when query has been loaded 1`] = `
             </div>
             <div
               data-primary="Contacts"
-              data-secondary="Permissions.numberOfPermissions 3"
+              data-secondary="3 permissions"
               data-testid="ListItemText"
             />
             <span
@@ -81,7 +81,7 @@ exports[`Permissions should display appName when query has been loaded 1`] = `
 </div>
 `;
 
-exports[`Permissions should display sorted app and konnector names when query is not loading 1`] = `
+exports[`AppList should display sorted app and konnector names when query is not loading 1`] = `
 <div>
   <div
     data-narrow="true"
@@ -90,7 +90,7 @@ exports[`Permissions should display sorted app and konnector names when query is
     <div
       data-testid="page-title"
     >
-      Permissions.title
+      Applications
     </div>
     <ul
       class="MuiList-root MuiList-padding"
@@ -114,7 +114,7 @@ exports[`Permissions should display sorted app and konnector names when query is
             </div>
             <div
               data-primary="Alan"
-              data-secondary="Permissions.numberOfPermissions"
+              data-secondary="0 permissions"
               data-testid="ListItemText"
             />
             <span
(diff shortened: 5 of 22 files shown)


src/components/App.jsx
```diff
@@ -44,16 +44,17 @@ export class App extends Component {
             <Route path="/connectedDevices/:deviceId" component={Devices} />
             <Route path="/sessions" component={Sessions} />
             <Route path="/storage" component={Storage} />
-            <Route exact path="/permissions" component={Permissions} />
+            <Route exact path="/permissions/apps" component={PermissionsTab} />
             <Route
               exact
-              path="/permissions/:app"
+              path="/permissions/apps/:app"
```
Crash--, line 50: if the route will also be used for konnector, please don't use the word `apps`. `slug` or something similar will be better

src/components/App.jsx
```diff
@@ -44,16 +44,17 @@ export class App extends Component {
...
-              path="/permissions/:app"
+              path="/permissions/slug/:app"
               component={PermissionsApplication}
             />
             <Route
-              path="/permissions/:app/:permission"
+              path="/permissions/slug/:app/:permission"
               component={Permission}
             />
+            <Route exact path="/permissions/data" component={PermissionsTab} />
```
trollepierre, line 57: why do we have 2 routes for this component?

trollepierre (commented):
first part (sorry, i got interrupt)

src/components/Sidebar.jsx
```diff
@@ -65,7 +65,7 @@ export const Sidebar = () => {
         {flag('settings.permissions-dashboard') && (
           <NavItem>
             <RouterLink
-              to="/permissions"
+              to="/permissions/slug"
```
trollepierre, line 68: consider using routes

src/containers/PermissionsApplication.jsx
```diff
@@ -87,7 +87,7 @@ const PermissionsApplication = ({ match, t }) => {
           <PageTitle>{appName.toUpperCase()}</PageTitle>
           {sortPermissionsByName(matchingQueryResult).map(
             ({ name, title, verbs }) => (
-              <Link to={`/permissions/${appName}/${name}`} key={name}>
+              <Link to={`/permissions/slug/${appName}/${name}`} key={name}>
```
trollepierre, line 90: consider using route

src/locales/en.json
```diff
@@ -402,6 +402,8 @@
     "read": "Read",
     "readAndWrite": "Read and Write",
     "write": "Write",
-    "numberOfPermissions" : "%{smart_count} permission |||| %{smart_count} permissions"
+    "numberOfPermissions" : "%{smart_count} permission |||| %{smart_count} permissions",
+    "applications": "Applicatiions",
```
trollepierre, line 406:
```suggestion
    "applications": "Applications",
```

trollepierre (commented):
second part

src/components/Permissions/DataList.spec.jsx
```diff
@@ -0,0 +1,37 @@
...
+  )
+})
+
+jest.mock('components/PageTitle', () => {
+  // eslint-disable-next-line react/display-name
+  return ({ children }) => <div data-testid="page-title">{children}</div>
+})
+
+describe('DataList', () => {
+  it('should display appName', () => {
```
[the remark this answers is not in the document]
Noemie21: I added a test to check that I display the right title

trollepierre (approved):

src/components/Permissions/DataList.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import Page from 'components/Page'
+import PageTitle from 'components/PageTitle'
+import { useI18n } from 'cozy-ui/transpiled/react'
```
Crash--, line 4: You should not import from this "root" directory. You should always use the deeper import you can for cozy-ui (in border to get lighter bundler at the end).

import { useI18n } from 'cozy-ui/transpiled/react/I18n'

src/components/Permissions/PermissionsTab.jsx
```diff
@@ -0,0 +1,45 @@
...
+function a11yProps(index) {
+  return {
+    id: `simple-tab-${index}`,
+    'aria-controls': `simple-tabpanel-${index}`
+  }
+}
+
+const PermissionsTab = () => {
+  const { t } = useI18n()
+  const [tab, setTab] = useState('apps')
```
Crash--, line 19: You should read the value from the router. Because here, you duplicate values from props to state but you also lose the information of the first URL I open. If I click on foo.cozy.cloud/permissions/data, the tab 'apps' will be selected but I don't want that, I want the data tabs. 

The handleChange method should not call a setState but should rather call an history.push()

src/components/Permissions/PermissionsTab.jsx
```diff
@@ -0,0 +1,45 @@
...
+        <Tab
+          value="data"
+          label={t('Permissions.data')}
+          href={`#${routes.dataList}`}
+          {...a11yProps(1)}
+        />
+      </Tabs>
+      <Divider className="u-mb-1" />
+      {tab === 'apps' ? <AppList /> : null}
+      {tab === 'data' ? <DataList /> : null}
```
Crash--, line 40: Can't you use 
```
<TabPanel index="apps"><AppList /><TabPanel>
<TabPanel index="data"><DataList /><TabPanel>
```
or something like that?

src/containers/Permission.jsx
```diff
@@ -73,20 +74,18 @@ const Permission = ({ match }) => {
...
-          >
-            <Icon icon={PreviousIcon} size={16} />
-          </IconButton>
-
+        <React.Fragment>
+          <a href={`#${routes.appList}/${appName.toLowerCase()}`}>
+            <IconButton className="u-mr-half">
+              <Icon icon={PreviousIcon} size={16} />
+            </IconButton>
+          </a>
```
Crash--, line 82: It's weird to have an `<a>`. 

Can't you use https://docs.cozy.io/cozy-ui/react/#!/Buttons with icons + href props ? 🤔

src/components/Permissions/PermissionsTab.jsx
Noemie21: I was inspired by this code which seems more recent: [MuiTabs ](https://github.com/cozy/cozy-ui/tree/67485ca3a6db28dd4ba64d2c91188c21b00034ce/react/MuiTabs )(2021) || [TabPanel](https://github.com/cozy/cozy-ui/blob/67485ca3a6db28dd4ba64d2c91188c21b00034ce/react/README.md) (2020)

Crash--: I see the misunderstanding: I thought that mui was exporting a TabPanel component since they use it on their exemple: https://v4.mui.com/components/tabs/#tabs  : 
```jsx
<AppBar position="static">
  <Tabs value={value} onChange={handleChange} aria-label="simple tabs example">
    <Tab label="Item One" {...a11yProps(0)} />
    <Tab label="Item Two" {...a11yProps(1)} />
    <Tab label="Item Three" {...a11yProps(2)} />
  </Tabs>
</AppBar>
<TabPanel value={value} index={0}>
  Item One
</TabPanel>
<TabPanel value={value} index={1}>
  Item Two
</TabPanel>
<TabPanel value={value} index={2}>
  Item Three
</TabPanel>
```

But in fact when checking the full source, TabPanel is a custom component: 
```jsx
return (
    <div
      role="tabpanel"
      hidden={value !== index}
      id={`simple-tabpanel-${index}`}
      aria-labelledby={`simple-tab-${index}`}
      {...other}
    >
      {value === index && (
        <Box p={3}>
          <Typography>{children}</Typography>
        </Box>
      )}
    </div>
  );
```

I think we should use something like this component at least for A11y reason

src/components/Permissions/PermissionsTab.jsx
Noemie21: I don't understand which component I should use

src/components/Permissions/PermissionsTab.jsx
Noemie21: I don't understand which component I should use

Crash--:
```jsx
const TabPanel = props => {
const { children, value, index, ...other } = props;
return  ( 
    <div role="tabpanel" 
    hidden={value !== index} 
    id={`simple-tabpanel-${index}`} 
    aria-labelledby={`simple-tab-${index}`} {...other} > 
     {value === index && ( {children} )} 
    </div> );
}
```

```jsx
<Tabs value={tab} onChange={handleChange}>
        <Tab
          value="apps"
          label={t('Permissions.applications')}
          href={`#${routes.appList}`}
          {...a11yProps(0)}
        />
        <Tab
          value="data"
          label={t('Permissions.data')}
          href={`#${routes.dataList}`}
          {...a11yProps(1)}
        />
      </Tabs>
      <Divider className="u-mb-1" />
<TabPanel value={tab} index="apps">
 <AppList />
<TabPanel> 
<TabPanel value={tab} index="data">
    <DataList />
<TabPanel>
``

File: src/components/Permissions/DataList.spec.jsx [hunk 11]
```diff
@@ -0,0 +1,29 @@
+import { render } from '@testing-library/react'
+import React from 'react'
+import DataList from './DataList'
+
+jest.mock('cozy-ui/transpiled/react', () => {
+  return { useI18n: () => ({ t: x => x }) }
+})
```
Crash--, line 7: nit: Instead of mocking useI18n each time for each component, you can create a Provider for your test. We call it `AppLike`. 

No need to be complicated at first, you can just add a `<I18n>` provider https://github.com/cozy/cozy-banks/blob/master/test/AppLike.jsx#L16-L22 

and then return it https://github.com/cozy/cozy-banks/blob/master/test/AppLike.jsx#L36

And when you test your component, instead of doing: 
```
    const { queryByText, container } = render(<DataList />)
```

you can do: 
```
    const { queryByText, container } = render(<AppLike><DataList /></AppLike>)
```

With that, no need to mock use18n anymore. 

This pattern will be useful to mock cozy-client store and else. We have it almost in all our apps.

src/containers/Permission.jsx
```diff
@@ -76,7 +76,7 @@ const Permission = ({ match }) => {
         <div>
           <IconButton
             className="u-mr-half"
-            href={`#/permissions/${appName.toLowerCase()}`}
+            href={`#/permissions/slug/${appName.toLowerCase()}`}
```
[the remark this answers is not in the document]
Crash--: Not agreed with this comment. `a` or `href` directly on the button is exactly the same for the router. If we were talking about a `<Link>` from react-router-dom then yes, there is some context we can lose (see what Link does exactly https://github.com/remix-run/react-router/blob/67f345b3e70f9df987c4b1b33c774e3310ce4aeb/packages/react-router-dom/index.tsx#L204-L231) 

But there, `a` or `href` on the IconButton are exactly the same for the router, but for the UI, this is not the same... So let's keep the use of IconButton with the href props

src/components/Permissions/PermissionsTab.jsx
Noemie21: Thank you

src/components/App.jsx
```diff
@@ -18,7 +18,7 @@ import Devices from 'containers/Devices'
 import Sessions from 'containers/Sessions'
 import Storage from 'containers/Storage'
 import Passphrase from 'containers/Passphrase'
-import Permissions from 'containers/Permissions'
+import PermissionsTab from './Permissions/PermissionsTab'
```
trollepierre, line 21:
```suggestion
import PermissionsTab from 'components/Permissions/PermissionsTab'
```

src/components/Permissions/DataList.spec.jsx
```diff
@@ -0,0 +1,34 @@
+import { render } from '@testing-library/react'
+import React from 'react'
+import DataList from './DataList'
+import AppLike from 'test/AppLike'
+
+jest.mock('cozy-ui/transpiled/react', () => {
+  return { useI18n: () => ({ t: x => x }) }
```
trollepierre, line 7: this should not be needed

src/components/Permissions/DataList.spec.jsx
```diff
@@ -0,0 +1,34 @@
...
+})
+
+describe('DataList', () => {
+  it('should display DataList title', () => {
+    const { queryByText, container } = render(
+      <AppLike>
+        <DataList />
+      </AppLike>
+    )
+    expect(container).toMatchSnapshot()
```
trollepierre, line 31: this is not needed

src/components/Permissions/DataList.spec.jsx
```diff
@@ -0,0 +1,34 @@
...
+    expect(queryByText('Data')).toBeTruthy()
```
trollepierre, line 32: according to Kent C Dodds: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen
```suggestion
    expect(getByText('Data')).toBeTruthy()
```

src/components/App.jsx
```diff
@@ -44,18 +44,19 @@ export class App extends Component {
             <Route path="/connectedDevices/:deviceId" component={Devices} />
             <Route path="/sessions" component={Sessions} />
             <Route path="/storage" component={Storage} />
-            <Route exact path="/permissions" component={Permissions} />
+            <Route exact path="/permissions/:page" component={PermissionsTab} />
             <Route
               exact
-              path="/permissions/:app"
+              path="/permissions/slug/:app"
```
Crash--, line 50: please don't use `app`. `slug` is the word to use 🙏

src/components/Permissions/PermissionsTab.jsx
```diff
@@ -0,0 +1,63 @@
...
+      {...other}
+    >
+      {value === index && children}
+    </div>
+  )
+}
+const PermissionsTab = ({ match, t }) => {
+  const slugOrData = match.params.page
+  const [tab, setTab] = useState(slugOrData)
+  const handleChange = (event, value) => setTab(value)
```
Crash--, line 34: @Noemie21 I'm still not happy with this implementation. We're copying a props to a state. We're creating duplication. And who says duplication, says synchronization problem. 

Can't we do something like that: 

```jsx
const PermissionsTab = () => {

const { t } = useI18n()
const { page } = useParams() // useParams is a hook coming from react-router 

const history = useHistory() // useHistory is a hook coming from react-router

const handleChange = (event, value) => history.push('/path/to/route/with/value)

return (
    <Page narrow>
      <Tabs value={page} onChange={handleChange}>
        <Tab
          value="slug"
          label={t('Permissions.applications')}
          href={`#${routes.appList}`}
          {...a11yProps(0)}
        />
        <Tab
          value="data"
          label={t('Permissions.data')}
          href={`#${routes.dataList}`}
          {...a11yProps(1)}
        />
      </Tabs>
      <Divider className="u-mb-1" />
      <TabPanel value={page} index="slug">
        <AppList />
      </TabPanel>
      <TabPanel value={page} index="data">
        <DataList />
      </TabPanel>
    </Page>
  )
}
```

Crash--: Like that, no more state involved, but you also get the back button of the browser working since you also change the URL.

src/components/Permissions/__snapshots__/AppList.spec.jsx.snap
```diff
@@ -33,7 +33,7 @@ exports[`Permissions should display appName when query has been loaded 1`] = `
             </div>
             <div
               data-primary="Alan"
-              data-secondary="Permissions.numberOfPermissions"
+              data-secondary="0 permissions"
```
Crash--, line 36: Is it wanted that 0 take an `s`?

src/components/Permissions/PermissionsTab.jsx
Noemie21: I replaced 

`const history = useHistory() // useHistory is a hook coming from react-router
const handleChange = (event, value) => history.push('/path/to/route/with/value)`

with 
``` const navigate = useNavigate()
  const handleChange = (event, value) => navigate(`/permissions/${value}`)```

is that ok ?

Crash--: Oh I didn't know about useNavigate()! Thanks

I see... This is a new API from react-router 6! Nice. 👏

Crash--: You get the idea about not using state since we don't need it. After that, the implementation is up to you. You are in a better position than me to write it ;)

PR #441, opened 2022-08-12
fix: Remove markdown on revokeDevice

acezard:
We had an issue where links could be inputs as device names (XSS).
We're removing markdown interpretation.
A side effect of this fix is that we lose the bold text on device name


src/components/DevicesModaleRevokeView.jsx
```diff
@@ -47,9 +48,10 @@ const RevokeDeviceDialog = ({ device, cancelAction, onDeviceRevoked }) => {
       onClose={cancelAction}
       content={
         <>
-          <ReactMarkdown
```
trollepierre, line 50: line 2 is not necessary, then ?

trollepierre (approved):

src/components/DevicesModaleRevokeView.jsx
acezard: it's still used L60

PR #442, opened 2022-08-12
fix(dependencies): react should not be a devDependencies

trollepierre:
The 3 commits are self explicit.

about eslint plugin import, I think it should be added in cozy-scripts (once accepted by the team)


acezard (approved):

zatteo:
Not the case anymore

PR #447, opened 2022-08-17
fix: Import from the deeper possible path

Crash--:
(I'll update the description and the commit once bundlemon will gave us the result)


Crash--:
Interesting that there is no diff at all. Will check later if there is no `cozy-*` dep that imports from the root.

trollepierre:
waiting for PR description

PR #459, opened 2022-08-31
feat: Flagship App Menu

acezard:
## Feature

I added a menu entry only visible on Flagship Apps or with a special flag. This entry allows the user to access the LockScreen page which contains its own menu. See [ticket](https://trello.com/c/HvxNN9eb/575-%E2%9A%99%EF%B8%8F-set-section-destin%C3%A9-%C3%A0-lapp-lorsque-jouvre-settings-depuis-laa-jh-2j) for more info.

## Context

This is a part of an important feature for Flagship App which allows users to lock their apps with fingerprints, or after a set time, etc. The whole feature is not finished with this ticket, it's only a part. See [ticket](https://trello.com/c/HvxNN9eb/575-%E2%9A%99%EF%B8%8F-set-section-destin%C3%A9-%C3%A0-lapp-lorsque-jouvre-settings-depuis-laa-jh-2j) for more info.

## Implementation

Added a LockScreen page, and added a link to this page. Every new implementation was done in TypeScript.
The PR also contains multiple small improvements to different parts of the app (see commits).

## Testing

No specific testing was added, only made sure no regression was introduced.

## Screenshots

image 
image

## Thoughts

There subsists some UI anomalies that are apparently linked to `cozy-bar`. These must be dealt with in a later PR and before the whole feature is finished.


src/components/Sidebar.tsx
```diff
@@ -27,6 +28,17 @@ export const Sidebar = (): JSX.Element => {
   return (
     <nav role="navigation">
       <List>
+        {isFlagshipApp() ||
+          (flag('settings.flagship-mode') && (
+            <MenuList title={t('Nav.header_flagship')}>
+              <MenuItemNavLink
+                to={routes.lockScreen}
+                primary={t('Nav.primary_lock_screen')}
+                icon={PeopleIcon} // @todo icon
```
trollepierre, line 37: is the PR ready for review?

acezard: yes I just need an icon to be merged in cozy-ui but everything is review ok

src/lib/flags.js
```diff
@@ -19,6 +19,7 @@ export const initFlags = () => {
 
 const flagsList = () => {
   flag('switcher', true)
-  flag('settings.partial-desktop-sync.show-synced-folders-selection'),
-    flag('settings.permissions-dashboard', true)
+  flag('settings.partial-desktop-sync.show-synced-folders-selection')
+  flag('settings.permissions-dashboard', true)
+  flag('settings.flagship-mode')
```
trollepierre, line 24: could it be unified?

flag('name', true) or flag('name') 🙏

src/components/Sidebar.tsx
acezard: @trollepierre https://github.com/cozy/cozy-ui/pull/2214

src/lib/flags.js
acezard: not sure what you mean sorry? some flags we don't want to activate them from the front-end, but from the stack

src/components/PageTitle.jsx
```diff
@@ -25,8 +25,8 @@ const PageTitle = ({ children, ...rest }) => {
     <>
       {!isRoot && (
         <BarLeft>
-          <IconButton onClick={navigateBack} size="large">
-            <Icon color="secondary" icon={LeftIcon} />
+          <IconButton onClick={navigateBack} size="large" title="Previous">
```
Crash--, line 28: Nice for the A11y! 

nit: should be translated ;)

src/lib/flags.js
Crash--: Yep, there is a different meaning between one arg vs 2 args: 
https://github.com/cozy/cozy-libs/blob/master/packages/cozy-flags/src/flag.js#L9-L17

One is a getter, two is a setter. 

It looks that we're setting `   flag('settings.permissions-dashboard', true) ` this flag to true, so if we publish a new version of the Settings app, by default everyone will have access to this page. I think this is not what we want (cc @Noemie21)

acezard: I can't speak for the other flags, I just know that flagship-mode must definitely not be enabled by defualt

src/components/PageTitle.jsx
acezard: updated

Crash--:
@acezard I appreciate the effort on the PR's description and the breakdown of the commits. There is still room for improvement, but this is already a very big step forward!

acezard:
> @acezard I appreciate the effort on the PR's description and the breakdown of the commits. There is still room for improvement, but this is already a very big step forward!

thanks, appreciated. 

Do you think we can approve it by the way?

src/lib/flags.js
trollepierre: > One is a getter, two is a setter.

In that case, we want to have a setter, then.

I think we should do:
```
flag('settings.permissions-dashboard', false)
  flag('settings.flagship-mode', false)
```

acezard: This would disable the flags coming from the stack I think?
flag('anything') will be false by default unless activated in features

trollepierre (approved):

src/lib/flags.js
trollepierre: ok, i start to understand.

The syntax of this feature is very confusing 😠

trollepierre (approved):

src/lib/flags.js
Noemie21: It's merged : https://github.com/cozy/cozy-settings/pull/462

PR #462, opened 2022-09-01
fix: Disable permissions flag

Noemie21:
https://github.com/cozy/cozy-settings/pull/459#discussion_r959429747

### DIFF
--- a/src/lib/flags.js
+++ b/src/lib/flags.js
@@ -20,6 +20,6 @@ export const initFlags = () => {
 const flagsList = () => {
   flag('switcher', true)
   flag('settings.partial-desktop-sync.show-synced-folders-selection')
-  flag('settings.permissions-dashboard', true)
+  flag('settings.permissions-dashboard')
   flag('settings.flagship-mode')
 }


src/lib/flags.js
```diff
@@ -20,6 +20,6 @@ export const initFlags = () => {
 const flagsList = () => {
   flag('switcher', true)
   flag('settings.partial-desktop-sync.show-synced-folders-selection')
-  flag('settings.permissions-dashboard', true)
-  flag('settings.flagship-mode')
+  flag('settings.permissions-dashboard', false)
+  flag('settings.flagship-mode', false)
```
trollepierre, line 24: No:
https://github.com/cozy/cozy-settings/pull/459#discussion_r960337471

1) please handle only `flag('settings.permissions-dashboard', true)`

2) i think we should use the value from cozy-stack `flag('settings.permissions-dashboard')`

trollepierre (changes requested):
see message.

ps: this PR needs to be tested with a big attention after the merge. Don't forget that, ask for help if needed

src/lib/flags.js
Noemie21: so I let `'settings.flagship-mode'` be as it was before?

Noemie21: @trollepierre ?

acezard: yes, it should be `flag('settings.flagship-mode')` here, so it can be activated from the Features panel (then forwarded by the stack to the front).

I'm not sure what will happen with the hardcoded `false`, might disable entirely the flag no matter what the stack says

trollepierre (approved):

PR #463, opened 2022-09-05
Don't use stack route to check if passphrase is valid

JF-Cozy:


src/locales/en.json
```diff
@@ -27,7 +27,7 @@
           "show": "Show",
           "required": "This field is required",
           "errors": {
-            "403": "Your password doesn't seem to be correct",
+            "passphrase": "Your password doesn't seem to be correct",
```
Ldoppea, line 30: question & nit: now that we are sure about the error type (i.e. "invalid password"). Should we continue to use "doesn't seem to be" instead of "is"?

Ldoppea (approved):
I reviewed the code, but not the `Add missing translation for cozy removing` commit as I don't speak those languages

Also there is a small typo in this commit message: `missions` instead of `missing`

src/locales/en.json
JF-Cozy: oui c'était le message initial pour la stack, qui était sûre elle aussi.

JF-Cozy:
> I reviewed the code, but not the Add missing translation for cozy removing commit as I don't speak those languages

tu peux me faire confiance, je parle couramment toutes ces langues... 🙄

PR #464, opened 2022-09-05
feat: Implement call to activate Biometry

acezard:
### Feature

This allows a user to activate or deactivate his app to be locked using biometry input (fingerprint etc.). Please see [this ticket](https://trello.com/c/BELZvEgc/576-%F0%9F%91%87-biometrie-activer-loption-biom%C3%A9trie-en-fake-jh-1-2j)

### Context

This fits into the broader Epic feature of activating biometry interactions on the Flagship App. The goal here is to implement step by step so multiple PRs are needed, or it would be impossible to review.

### Implementation

Just use an async call to WebviewIntent and the response to activating/deactivating the UI setting. Also checking injected values at Webview init. Also created new components to handle these features.

### Testing

No specific testing was done as I'm waiting for more real implementations to be here (like fingerprint input).

### Screenshots

image

### Thoughts

Not much to say here it's pretty straightforward. Maybe need to create a sort of hook or something to handle all the future settings in a DRY way.


src/components/pages/LockScreen.tsx
```diff
@@ -23,6 +44,10 @@ export const LockScreen = (): JSX.Element => {
           <MenuItemSwitch
             primary={t('Nav.primary_biometry')}
             icon={Fingerprint}
+            onClick={(): void =>
+              void handleChange(setHasBiometry, webviewIntent)
+            }
+            checked={hasBiometry}
```
Ldoppea, line 50: nit: got a bit confused by `hasBiometry` name. I expected it to contain the "capability" and not the fact the user enabled it. Maybe `shouldUnlockWithBiometry`?

Ldoppea (approved):

src/components/pages/LockScreen.tsx
acezard: I'll recheck naming convention when more settings are added but you're right

PR #470, opened 2022-09-07
fix: Remove flag for Subscription link

acezard:
It is not needed as the instance context already knows if the feature
is enabled or not.
It probably was a legacy check for testing purpose.
Also fixed a typing
Also updated the button locale in EN

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -76,7 +76,7 @@
     "primary_faq": "Manuel d'utilisation",
     "primary_lock_screen": "Unlock screen",
     "primary_logout": "Déconnexion",
-    "primary_package": "Forfait",
+    "primary_package": "Subscription",
     "primary_photo_sync": "Synchronisation auto des photos",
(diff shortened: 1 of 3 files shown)


src/locales/en.json
```diff
@@ -76,7 +76,7 @@
     "primary_faq": "Manuel d'utilisation",
     "primary_lock_screen": "Unlock screen",
     "primary_logout": "Déconnexion",
```
Crash--, line 78: there is a lot of french word in this feature. Is it intended?

acezard: Not really, I don't have the translations at this point unless I missed it

Crash--: ping @poupotte

acezard: @Crash-- we talked about it in a meeting, we'll review these keys/translations in another PR

Crash-- (approved):

PR #471, opened 2022-09-08
fix(Permissions): improve design using NavigationList

vdnj:
Capture d’écran 2022-09-08 à 15 02 42

### DIFF
--- a/src/components/Permissions/AppPermissions.jsx
+++ b/src/components/Permissions/AppPermissions.jsx
@@ -13,10 +13,8 @@ import {
 import Icon from 'cozy-ui/transpiled/react/Icon'
 import IconButton from 'cozy-ui/transpiled/react/IconButton'
 import PreviousIcon from 'cozy-ui/transpiled/react/Icons/Previous'
-import NavigationList, {
-  NavigationListSection
-} from 'cozy-ui/transpiled/react/NavigationList'
 import ListItemIcon, {
+  smallSize,
   mediumSize
 } from 'cozy-ui/transpiled/react/MuiCozyTheme/ListItemIcon'
 import ListItemText from 'cozy-ui/transpiled/react/ListItemText'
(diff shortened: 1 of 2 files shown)


src/components/Permissions/AppPermissions.jsx
```diff
@@ -93,6 +94,12 @@ const PermissionsApplication = ({ t }) => {
 
   const isKonnector = type => type === 'io.cozy.konnectors'
 
+  const isNotLastItem = name =>
+    name !==
+    sortPermissionsByName(matchingQueryResult)[
```
trollepierre, line 99: let's extract this in one single variable 🙏

vdnj: I don't understand what you mean. It is a function that returns either `true` or `false`

Crash--: You call several times `sortPermissionsByName` with the same arg `matchingQueryResult`

You should call only once `sortPermissionsByName(matchingQueryResult)`. You can do something like:

```
const sortedPermissionsByName = sortPermissionsByName(matchingQueryResult)
const isNotLastItem = name => {
    return (
      name !==
      sortedPermissionsByName[
        sortedPermissionsByName.length - 1
      ].name
    )
  }
  ```

and also there is in the component: 
```
{sortPermissionsByName(matchingQueryResult).map(
```

that you can change to: 
```
sortedPermissionsByName.map(
```

File: src/components/Permissions/AppPermissions.jsx [hunk 2]
```diff
@@ -13,10 +13,8 @@ import {
 import Icon from 'cozy-ui/transpiled/react/Icon'
 import IconButton from 'cozy-ui/transpiled/react/IconButton'
 import PreviousIcon from 'cozy-ui/transpiled/react/Icons/Previous'
-import NavigationList, {
-  NavigationListSection
-} from 'cozy-ui/transpiled/react/NavigationList'
 import ListItemIcon, {
+  smallSize,
```
trollepierre, line 17: there is typo in the commit message "Permissions" not "Pemrissions"

src/components/Permissions/AppPermissions.jsx
```diff
@@ -93,6 +94,15 @@ const PermissionsApplication = ({ t }) => {
 
   const isKonnector = type => type === 'io.cozy.konnectors'
 
+  const isNotLastItem = name => {
+    return (
+      name !==
+      sortPermissionsByName(matchingQueryResult)[
+        sortPermissionsByName(matchingQueryResult).length - 1
+      ].name
```
trollepierre, line 102: let's call to refactor this function 🙏

trollepierre (approved):
nit: commit naming is not so good, i would use the same name as the PR title I updated.

PR #472, opened 2022-09-08
fix: Use only one way to inject data

Crash--:
This is the recommended way in order to get all the new added attributes without editing the ejs.


JF-Cozy (approved):
changelog ? 👼

trollepierre:
there is no changelog in cozy-settings ✅

PR #473, opened 2022-09-09
fix: Remove vertical scrolling on mobile menu

acezard:
### Bug

Related to https://trello.com/c/FhQCY4DE/590-scroll-inutile-dans-le-menu-de-settings

When on the Menu page on mobile, there is an unwanted and apparently useless Y scrolling ability. It makes the UI look semi-broken.

### Fix

Overwriting the `:after` pseudo-element in `Layout` when on a small screen because this page is a special case and doesn't have anything to show on the bottom of the page (like a bottom sidebar which we do not have anymore on `cozy-settings`). So we just remove that pseudo-element which was creating unwanted scrolling.

### Testing

Manual testing in the browser

### Thoughts

At this point, I didn't think updating `cozy-ui` was necessary because it is a special app scenario

### DIFF
--- a/src/styles/index.styl
+++ b/src/styles/index.styl
@@ -4,3 +4,7 @@
     .set-markdown-link
         color var(--dodgerBlue)
         text-decoration none
+
+.Layout--small-view
+    &::after
+        content none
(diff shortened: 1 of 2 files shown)


src/styles/index.styl
```diff
@@ -4,3 +4,7 @@
     .set-markdown-link
         color var(--dodgerBlue)
         text-decoration none
+
+.Layout--small-view
+    &::after
+        content none
```
Crash--, line 10: @JF-Cozy FYI

Crash-- (approved):

PR #475, opened 2022-09-12
Feat: Implement auto lock setting

acezard:
### Feature

Trello ticket: https://trello.com/c/Rcw6Fxh7/578-%F0%9F%91%87-biometrie-pouvoir-verrouiller-lapp-version-on-charge-cozy-client-avant-unlock-jh-3-4j

As a user, I want to be able to toggle on/off the autolocking of my mobile application.

### Context

This is only the web part of the feature, it needs the native part to be implemented in order to work. It's a feature in the broader context of biometry/security features.

### Implementation

Following the implementation of the biometry setting, we have a controlled value at bootstrap coming from native, then at runtime, we listen to the returned value should we change the value with the switch.

### Testing

No tests were added, only manual testing. Tests will be added at a later date because the final API is not really decided yet. 

### Screenshots

image

### Thoughts

The current implementation is not super scalable and will have to be refactored in some way. Also, the input is not totally controlled, it's only controlled on the first render, then it's hooked to react-native with post-me in real-time.


src/components/menu/MenuList.tsx
```diff
@@ -1,31 +1,35 @@
...
-            .map((child, index, array) => (
-              <React.Fragment key={index}>
-                {child}
-                {index !== array.length - 1 && (
-                  <Divider component="li" variant="inset" />
-                )}
-              </React.Fragment>
-            ))
-        : children}
+      <li>
```
acezard, line 17: `<li />` are used because `<ListItem />` has too many specific styles that are cumbersome to remove.
The main reason to add these tags is to keep a semantic and valid HTML5.

trollepierre (approved):

src/locales/fr.json
```diff
@@ -75,13 +75,14 @@
     "primary_biometry": "Déverrouiller avec mes empreintes biométriques",
     "primary_faq": "Manuel d'utilisation",
     "primary_lock_screen": "Écran de verrouillage",
+    "primary_lock_switch": "Verrouiller en passant l’app en arrière plan",
```
Ldoppea, line 78: Wondering about this.

Our figma spec says `Vérrouiller en quittant l'app`
Our trello card says `Vérrouiller en quittant l'app ou en la passant en arrière plan`
This PR says `Verrouiller en passant l'app en arrière plan`

PR #478, opened 2022-09-14
Redesign DataPermissions

vdnj:
Capture d’écran 2022-09-13 à 16 34 54


src/locales/fr.json
```diff
@@ -430,7 +430,8 @@
     "numberOfApplications" : "%{smart_count} application |||| %{smart_count} applications",
     "applications": "Applications",
     "data": "Données",
-    "open": "Ouvrir"
+    "open": "Ouvrir",
+    "access": "Les applications disposant de cette permission peuvent accéder à votre/vos "
```
trollepierre, line 434: we need to use plural form using smart count (like line 430)

src/components/Permissions/DataPermissions.jsx
```diff
@@ -1,35 +1,40 @@
...
+import CozyClient, {
+  Q,
+  useQuery,
+  isQueryLoading,
+  hasQueryBeenLoaded
+} from 'cozy-client'
 import { APPS_DOCTYPE, KONNECTORS_DOCTYPE } from 'doctypes'
-import { Link, useParams } from 'react-router-dom'
-import withAllLocales from 'lib/withAllLocales'
+import { completePermission } from './DataListHelpers'
```
trollepierre: if it's used on DataPermissions, this is not a DataListHelpers

trollepierre: let's move this method inside a helper file in Permissions/helpers.js

src/components/Permissions/DataPermissions.spec.jsx
```diff
@@ -1,139 +0,0 @@
-import { render } from '@testing-library/react'
```
trollepierre, line 1: let's re-add this test

src/locales/fr.json
vdnj: I think smart_count counts the number of applications. This is not the same usage here. What I can try to do is checking if the last letter of the Permission is an "s". If this is an "s" then I write "vos" else I write "votre".

trollepierre: if our PM agrees to use votre/vos, let's keep it

trollepierre (approved):

PR #489, opened 2022-09-28
feat: Implement final Biometry setting

acezard:
### Feature

- [x] Greyed biometry activation if phone does not support it OR if it's not available
- [x] Wording depends on biometry type
- [x] Icon depends on biometry type

### Implementation

- Assuming "not available" covers cases where phone doesn't support biometry at all
- Simple mapping
- Simple mapping

### Testing

No unit tests were written, only manual testing. This is definitely not ideal in this case, the view is more complex than it should be.

### Screenshots

image

### Thoughts

It started simple, but there is now a LOT of logic in the `LockScreen` view. This is not acceptable in terms of technical debt. Refactoring is needed as soon as possible. It could not be done during this PR but there are two PR left on this feature, it would be a huge bonus.


src/components/menu/MenuItemNavLink.tsx
```diff
@@ -38,11 +38,9 @@ export const MenuItemNavLink = (props: {
   return (
     <li>
       <ListItem button component={renderLink}>
-        {icon ? (
```
Ldoppea, line 41: This seems to be more than just type improve

src/components/menu/MenuItemSwitch.tsx
```diff
@@ -13,20 +13,23 @@ interface MenuItemSwitchProps {
...
 }
 
 export const MenuItemSwitch = ({
   checked,
   icon,
   onClick,
   primary,
-  secondary
+  secondary,
+  disabled
```
Ldoppea, line 25: This seems to be more than just type improve

src/hooks/useDiskPercentage.ts
```diff
@@ -29,7 +31,7 @@ export const useDiskPercentage = (): string | undefined => {
       }
     }
 
-    !percent && client && asyncCore()
```
Ldoppea, line 32: This seems to be more than just type improve

src/hooks/useOffersLink.ts
```diff
@@ -39,7 +41,7 @@ export const useOffersLink = (): undefined | string | null => {
       }
     }
 
-    offersLink === undefined && client && asyncCore()
```
Ldoppea, line 42: This seems to be more than just type improve

Ldoppea (approved):

PR #495, opened 2022-10-11
feat: Implement PIN code setting

acezard:
### Feature

- [x] View: new modal to create and confirm PIN code
- [x] View logic: control new PIN setting from the native app
- [x] Business logic: allow the web app to store the new PIN code into the native app
- [x] Business logic: allow the web app to delete the existing PIN code from the native app

### Implementation

Most of the implementation was done in a new component named `PinCodeDialog`. It tries to stay as simple as possible and leverages React `useState` a lot.

### Screenshots

Screenshot_1665491746
Screenshot_1665492493
Screenshot_1665492515


src/components/pages/LockScreen.tsx
```diff
@@ -47,11 +54,24 @@ export const LockScreen = (): JSX.Element => {
   const onBiometryLock = (): void =>
     void handleChange(setBiometry, 'biometryLock', webviewIntent)
 
+  const onPinCodeLock = async (pinCode?: string): Promise<void> => {
+    if (!pinCodeEnabled && !pinCode) return setPinModalVisible(true)
```
Ldoppea, line 58: I don't understand how we can reach this code if `pinCodeEnabled` is `false`.

src/components/pages/LockScreen.tsx
```diff
@@ -73,13 +93,27 @@ export const LockScreen = (): JSX.Element => {
           />
 
           <MenuItemSwitch
-            primary={t('Nav.primary_lock_switch')}
+            checked={pinCodeEnabled}
+            icon={Password}
+            onClick={(): void => void onPinCodeLock()}
```
Ldoppea, line 98: Same warning as in previous PR, we should find a way to just put pointers on this prop. We don't want to declare an anonymous method just to satisfy typescript linter.

src/components/dialogs/PinCodeDialog.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const handleChange = (
+    event: ChangeEvent<HTMLInputElement> | KeyboardEvent
+  ): void => {
+    if ('value' in event.target) {
+      flowStep === 'initial'
+        ? setInitialValue(event.target.value)
+        : setConfirmValue(event.target.value)
+    }
+
+    if ('key' in event && flowStep === 'initial') {
```
Ldoppea, line 45: why do we check `'key' in event` here? We don't use it after. Is that a trick to detect if the change comes from keyboard input or something similar?

src/components/dialogs/PinCodeDialog.tsx
```diff
@@ -0,0 +1,140 @@
...
+      validatePinCode(initialValue) && setFlowStep('confirm')
+      inputRef.current?.blur()
+    }
+
+    if ('key' in event && flowStep === 'confirm')
+      validatePinCode(initialValue) && trySetPinCode()
+  }
+
+  const trySetPinCode = (): void => {
+    initialValue !== confirmValue
```
Ldoppea, line 55: nit: i'm not sure that it is a good thing to abuse usage of ternary operators when a simple if-else cab be used. I my opinion this there is no pros to use this form and it makes the code less maintainable for external developers.

src/components/dialogs/PinCodeDialog.tsx
```diff
@@ -0,0 +1,140 @@
...
+            }
+            label={t(
+              `LockScreenView.${
+                flowStep === 'initial' ? 'choose_pin_next' : 'confirm_pin_next'
+              }`
+            )}
+            disabled={
+              flowStep === 'initial'
+                ? initialValue.replace(/_/g, '').length < 4
+                : confirmValue.replace(/_/g, '').length < 4
```
Ldoppea, line 97: Maybe `validatePinCode` can be used here (even if it has a `===` instead of a `<`)?

src/locales/en.json
```diff
@@ -361,6 +362,16 @@
...
+    "choose_pin_body": "choose_pin_body",
+    "choose_pin_cancel": "choose_pin_cancel",
+    "choose_pin_label": "choose_pin_label",
+    "choose_pin_next": "choose_pin_next",
+    "choose_pin_title": "choose_pin_title",
+    "confirm_pin_body": "confirm_pin_body",
+    "confirm_pin_cancel": "confirm_pin_cancel",
+    "confirm_pin_error": "confirm_pin_error",
+    "confirm_pin_next": "confirm_pin_next",
+    "confirm_pin_title": "confirm_pin_title",
```
Ldoppea, line 374: Just a reminder that some english strings are not defined yet :)

Ldoppea (approved):
Everything looks good. My comments are more about clarifications than change requests.

Don't forget English translations.

nit: I'm wondering if PinCodeDialog can be simplified by handling steps in an upper component and having a generic pin dialog. Maybe this would avoid all those ternary operators (but i'm not sure about that)

src/components/pages/LockScreen.tsx
acezard: Well not sure to understand. If it's false it will be if (true && !pinCode) so we just want to be sure there is no pinCode argument so we can activate the pinCode dialog

src/components/pages/LockScreen.tsx
acezard: I'm still trying to find a way to convert onClick handler to an async call but TypeScript is smarter than me and always finds that I hid a promise somewhere in the callback.

src/components/dialogs/PinCodeDialog.tsx
acezard: Yes that allows us to know (and TypeScript too) that the event type is discriminated to be a keyboardinput here

src/components/dialogs/PinCodeDialog.tsx
acezard: Definitely, seems like I forgot to use it in fact

src/locales/en.json
acezard: Yes, we'll have the localisations at a later point I believe, I notified about it

acezard:
> Everything looks good. My comments are more about clarifications than change requests.
> 
> Don't forget English translations.
> 
> nit: I'm wondering if PinCodeDialog can be simplified by handling steps in an upper component and having a generic pin dialog. Maybe this would avoid all those ternary operators (but i'm not sure about that)

Yea I agree about that last point and actually wanted to refactor it but I ran out of time. All those ternary work but don't look that good. I'll see if I have some time at a later point

src/components/pages/LockScreen.tsx
Ldoppea: My point was that if `pinCodeEnabled=false` I don't get how `onPinCodeLock` can be called.

src/components/pages/LockScreen.tsx
Ldoppea: Maybe we should remove this specific rule then? I'm happy with all the TS checks, but I don't like the fact that we have to modify the app's behaviour just to pass them.

src/components/pages/LockScreen.tsx
acezard: sorry not sure to follow here, onPinCodeLock can be called with false or true for pinCodeEnabled

PR #498, opened 2022-10-14
fix: improve lockscreen consistency

acezard:
Locales, UX

### DIFF
--- a/src/components/pages/LockScreen.tsx
+++ b/src/components/pages/LockScreen.tsx
@@ -22,7 +22,7 @@ const handleChange = async (
   setting: 'biometryLock' | 'autoLock' | 'PINLock',
   webviewIntent?: WebviewService,
   params?: Record<string, unknown>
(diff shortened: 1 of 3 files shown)


src/components/pages/LockScreen.tsx
```diff
@@ -32,6 +32,8 @@ const handleChange = async (
     : logger.error(
         `Error while calling toggleSetting('${setting}'), returned value is null or undefined."`
       )
+
+  return res
```
Ldoppea, line 36: Shouldn't we rely on the `dispatch` here?

Ldoppea (approved):
Code seems to work, but there is a question about API consistency. Feel free to merge now and check the consistency issue later

src/components/pages/LockScreen.tsx
acezard: Initiallly yes, but for biometry I need the actual return value

PR #501, opened 2022-10-19
Feat/nav subscriptions

Crash--:
Should reduce the flickering in the flagship app. 🤞

### DIFF
--- a/src/hooks/useOffersLink.ts
+++ b/src/hooks/useOffersLink.ts
@@ -1,5 +1,5 @@
 import { useEffect, useState } from 'react'
-import { useClient, Q } from 'cozy-client'
+import CozyClient, { useClient, Q } from 'cozy-client'
 import logger from 'lib/logger'
 
 const warn = (logger as unknown as { warn: (...args: unknown[]) => void }).warn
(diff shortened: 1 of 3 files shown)


src/hooks/useOffersLink.ts
```diff
@@ -20,14 +20,20 @@ export const useOffersLink = (): undefined | string | null => {
...
-
-        const { data: instance } = (await client.query(
-          Q('io.cozy.settings').getById('instance')
-        )) as ExpectedInstance
-
+        const [{ data: context }, { data: instance }] = await Promise.all([
+          (await client.fetchQueryAndGetFromState({
+            definition: Q('io.cozy.settings').getById('context'),
+            options: {
+              fetchPolicy: CozyClient.fetchPolicies.olderThan(30 * 100000)
```
Ldoppea, line 27: We often read in our codebase `olderThan(30 * 1000)` which is easily understandable as `30 x 1s` because we are used to manipulate milliseconds.

Here you used `30 * 100000` which may be confusing. A reader may wonder what is the unit here (I was going to give microsecond as exemple but then I noticed there were only 5 zeros 😅 )

I would prefer `3000 * 1000` or better : a named constant

Crash--: I would prefer https://github.com/cozy/cozy-client/issues/1253 :)

Crash--: (force pushed and changed to `3000 * 1000`)

PR #502, opened 2022-10-19
Fix/mistakes

Crash--:


paultranvan (approved):
Always better when working for the actual code :D

PR #503, opened 2022-10-19
refactor: No need to use CozyClient.fromDOM

Crash--:
Let's update this old code to follow new convention. Also fix a warning 'CozyClient is already logged'


paultranvan (approved):
Nice cleanup :clap:

PR #505, opened 2022-10-20
fix: use singleDocData

Crash--:
When refactored query() to fetchQueryAndGetFromState() we lose this option, we fixed the real issue in cozy-client https://github.com/cozy/cozy-client/commit/a840bf46e2d6b5f28572e3f1d9f9558765d8e085 but let's be explict, this is better

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
   "dependencies": {
     "@material-ui/lab": "4.0.0-alpha.61",
     "classnames": "2.3.1",
-    "cozy-client": "^33.1.2",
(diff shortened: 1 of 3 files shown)


Merkur39 (approved):

package.json
```diff
@@ -62,6 +62,7 @@
     "cozy-keys-lib": "^3.11.1",
     "cozy-realtime": "3.13.1",
     "cozy-sharing": "3.10.0",
+    "cozy-stack-client": "^33.2.0",
```
paultranvan, line 65: Why did you need to add this?

PR #507, opened 2022-10-27
feat: Upgrade cozy-client to get remote-doctypes locales

vdnj:
Added doctypes translation in cozy-client


paultranvan:
Could you rephrase a bit the commit to explain why this upgrade? Something like `feat: Upgrade cozy-client to get remote-doctypes locales`

PR #509, opened 2022-11-02
feat: Filter 'last outgoing data history' section elements

vdnj:
- Before, all remote requests elements were shown in 'Last outgoing data history' section, now only items created by the current app are shown
- Before "Monthly Stats" was the name of all remote requests elements, now it is the name of the doctype
- Renamed this section's file name 
Capture d’écran 2022-11-02 à 16 42 32

### DIFF
--- a/src/components/Permissions/AppPermissions/AppPermissions.jsx
+++ b/src/components/Permissions/AppPermissions/AppPermissions.jsx
@@ -10,9 +10,9 @@ import AppIcon from 'cozy-ui/transpiled/react/AppIcon'
 import Icon from 'cozy-ui/transpiled/react/Icon'
 import IconButton from 'cozy-ui/transpiled/react/IconButton'
 import PreviousIcon from 'cozy-ui/transpiled/react/Icons/Previous'
-import { OpenappButton } from '../OpenappButton'
-import { AboutButton } from '../AboutButton'
-import withAllLocales from '../../../lib/withAllLocales'
+import { OpenappButton } from 'components/Permissions/OpenappButton'
+import { AboutButton } from 'components/Permissions/AboutButton'
+import withAllLocales from 'lib/withAllLocales'
 import NavigationList from 'cozy-ui/transpiled/react/NavigationList'
 import CozyClient, {
   Q,
@@ -20,13 +20,14 @@ import CozyClient, {
   isQueryLoading,
   hasQueryBeenLoaded
 } from 'cozy-client'
-import AccessRightsSection from '../AccessRightsSection'
-import LatestDataReleases from '../LatestDataReleases'
+import AccessRightsSection from 'components/Permissions/AccessRightsSection'
+import LatestOutgoingDataHistory from 'components/Permissions/LatestOutgoingDataHistory'
 import {
   filterPermissions,
   sortPermissionsByName,
-  completeAppPermission
-} from '../helpers/permissionsHelper'
+  completeAppPermission,
+  filterRemoteRequests
+} from 'components/Permissions/helpers/permissionsHelper'
 import useFetchJSON from 'cozy-client/dist/hooks/useFetchJSON'
 
 const AppPermissions = ({ t }) => {
@@ -86,6 +87,11 @@ const AppPermissions = ({ t }) => {
   )
 
   const isKonnector = type => type === 'io.cozy.konnectors'
+
+  const filteredRemoteRequests = filterRemoteRequests(
+    queryResultRemoteRequests,
+    slugName
+  )
 
   return (
     <Page narrow>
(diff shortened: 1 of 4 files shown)


src/components/Permissions/helpers/permissionsHelper.js
```diff
@@ -106,3 +106,13 @@ export const sortDataByDate = queryResult => {
     (a, b) => new Date(b.created_at) - new Date(a.created_at)
   )
 }
+
+export const filterRemoteRequests = (remoteRequests, slugName) => {
+  const filteredRemoteRequests = { data: [] }
```
Ldoppea, line 111: Maybe you can get rid of the `.data` complexity here and just return a plain array

src/components/Permissions/helpers/permissionsHelper.js
```diff
@@ -106,3 +106,13 @@ export const sortDataByDate = queryResult => {
...
+  for (let i = 0; i < remoteRequests.data.length; i++) {
+    if (remoteRequests.data[i].cozyMetadata.createdByApp === slugName) {
+      filteredRemoteRequests.data.push(remoteRequests.data[i])
+    }
+  }
```
Ldoppea, line 116: Here you can use `Array.filter` instead of a `for` loop. Result would be the same but the code would comply with modern conventions. Also this reduce the risk of error when manipulating indexes.

```js
filteredRemoteRequests.data = remoteRequests.data.filter(data => cozyMetadata.createdByApp === slugName)
```

And if you choose to remove the `.data` complexity as suggested in previous comment, you can just use:

```js
return remoteRequests.data.filter(data => cozyMetadata.createdByApp === slugName)
```

src/components/Permissions/AppPermissions/AppPermissions.jsx
```diff
@@ -87,6 +88,10 @@ const AppPermissions = ({ t }) => {
 
   const isKonnector = type => type === 'io.cozy.konnectors'
 
+  const filteredRemoteRequests =
+    queryResultRemoteRequests.data?.length > 0 &&
```
Ldoppea, line 92: nit: Here you can move the `length` check in the filter method. This would make the method more secure if you choose to re-use it somewhere else in the future.

Ldoppea (approved):
I approve the code as it should work correctly. However I suggested some code style improvements. Feel free to implement them or not.

src/components/Permissions/AppPermissions/AppPermissions.jsx
```diff
@@ -21,11 +21,12 @@ import CozyClient, {
   hasQueryBeenLoaded
 } from 'cozy-client'
 import AccessRightsSection from '../AccessRightsSection'
-import LatestDataReleases from '../LatestDataReleases'
+import LatestDataReleases from '../LatestOutgoingDataHistory'
```
Crash--, line 24: You should never use relative import. 🙏

src/components/Permissions/AppPermissions/AppPermissions.jsx
```diff
@@ -147,9 +152,9 @@ const AppPermissions = ({ t }) => {
               />
             )}
 
-            {queryResultRemoteRequests.data.length > 0 && (
+            {filteredRemoteRequests.data?.length > 0 && (
```
Crash--, line 155: Since you currently instanciate `filteredRemoteRequests` by doing : 
```
const filteredRemoteRequests =
    queryResultRemoteRequests.data?.length > 0 && ...
```

The variable can be false. So it's weird to have: `false.data?.length`.

src/components/Permissions/helpers/permissionsHelper.js
vdnj: I did it because the function `sortDataByDate` used this format. I changed both function so it's cleaner now

src/components/Permissions/AppPermissions/AppPermissions.jsx
```diff
@@ -10,23 +10,24 @@ import AppIcon from 'cozy-ui/transpiled/react/AppIcon'
...
 import CozyClient, {
   Q,
   useQuery,
   isQueryLoading,
   hasQueryBeenLoaded
 } from 'cozy-client'
-import AccessRightsSection from '../AccessRightsSection'
-import LatestDataReleases from '../LatestDataReleases'
+import AccessRightsSection from 'components/Permissions/AccessRightsSection'
+import LatestDataReleases from 'components/Permissions/LatestOutgoingDataHistory'
```
Ldoppea, line 24: I just noticed that `LatestDataReleases` name mismatches the `LatestOutgoingDataHistory` component name. Is that intended?

Ldoppea (approved):

PR #511, opened 2022-11-07
fix: Remove is not last item function

vdnj:
This function is not useful. Improved code by removing it and checking with the index.


Ldoppea (approved):
Nice commit messages 👍

PR #512, opened 2022-11-07
feat: Handle denied biometry access on iOS

Ldoppea:
On iOS with FaceID the biometry feature is guarded by OS level authorizations

If the user reject the authorisation process, then the application won't be able to use FaceID anymore until the user re-authorizes it from the OS settings

We want this information to be available from the settings cozy-app and to be able to redirect the user to the OS settings
so they can re-autorize the biometry

___

### Related PRs:

- https://github.com/cozy/cozy-libs/pull/1870
- https://github.com/cozy/cozy-flagship-app/pull/457

___

### TODO:

- [x] Update cozy-intent and cozy-device-helper after https://github.com/cozy/cozy-libs/pull/1870 being merged
- [x] Replace text/translation with final values


src/components/dialogs/BiometryDeniedDialog.tsx
```diff
@@ -0,0 +1,56 @@
...
+            label={t('BiometryDeniedDialog.open_settings')}
+          />
+        </>
+      }
+      content={
+        <>
+          <Typography className="u-mb-1" variant="body1">
+            {t('BiometryDeniedDialog.content')}
+          </Typography>
+        </>
```
Crash--, line 52: Nit: I think it’s the first time I’be to read actions before the content. You shifted my mental model here 😂

src/locales/en.json
```diff
@@ -76,12 +76,13 @@
     "primary_biometry_android": "Unlock with my biometric fingerprints",
     "primary_biometry_faceid": "Unlock with my Face ID",
     "primary_biometry_touchid": "Unlock with my Touch ID",
+    "primary_biometry_denied_title": "Unlock with my biometric fingerprints",
```
Crash--, line 79: Is it still needed?

Crash-- (approved):

src/locales/en.json
Ldoppea: Good catch, was from a first attempt before choosing to create the `BiometryDeniedDialog` section. Fixed

src/components/dialogs/BiometryDeniedDialog.tsx
Ldoppea: This is because you didn't reviewed https://github.com/cozy/cozy-settings/pull/495 :trollface: 

I did a copy-paste of the PinCodeDialog component as a basis and I didn't even notice the ordering. I reversed the order

PR #513, opened 2022-11-08
fix: Add overlay to UI when using biometry

acezard:
### Feature

Biometry settings on mobile app

### Context

The biometry input wasn't triggering any bar overlay change, they stayed white

### Implementation

Adding (for Android only as iOS doesn't have the issues) setFlagshipUI calls on open/close of the biometry prompt

### Testing

Manual only

### Screenshots

image

### Thoughts

Code is verbose as is usual with setFlagshipUi but this topic can't be tackled in a bug fix

### DIFF
--- a/src/components/pages/LockScreen.tsx
+++ b/src/components/pages/LockScreen.tsx
@@ -74,15 +74,33 @@ export const LockScreen = (): JSX.Element => {
         return setBiometryDeniedDialogVisible(true)
       }
 
+      if (!biometryEnabled && flagshipMetadata.platform?.OS === 'android')
+        await webviewIntent.call('setFlagshipUI', {
+          bottomOverlay: 'rgba(0,0,0,0.5)',
+          topOverlay: 'rgba(0,0,0,0.5)'
+        })
+
       const value = await handleChange(
         setBiometry,
         'biometryLock',
         webviewIntent
       )
+
       value && setAutoLock(true)
+
+      if (!biometryEnabled && flagshipMetadata.platform?.OS === 'android')
+        await webviewIntent.call('setFlagshipUI', {
+          bottomOverlay: 'transparent',
+          topOverlay: 'transparent'
+        })
     }
 
-    void doOnBiometryLock()
+    doOnBiometryLock().catch(onrejected =>
+      logger.error(
+        `Error while calling onBiometryLock(), see error:`,
+        onrejected
+      )
+    )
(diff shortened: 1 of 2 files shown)


src/components/pages/LockScreen.tsx
```diff
@@ -53,10 +53,46 @@ export const LockScreen = (): JSX.Element => {
   const biometryAvailable = Boolean(flagshipMetadata.biometry_available)
   const biometryType = flagshipMetadata.biometry_type
 
-  const onBiometryLock = (): void =>
-    void handleChange(setBiometry, 'biometryLock', webviewIntent).then(
-      value => value && setAutoLock(true)
-    )
+  const onBiometryLock = (): void => {
```
Crash--, line 56: we can't use an async method here?

src/components/pages/LockScreen.tsx
```diff
@@ -53,10 +53,46 @@ export const LockScreen = (): JSX.Element => {
...
+    if (!webviewIntent) return
```
Crash--, line 57: nit: Do we really need this check here? Shouldn't be done before within the Settings' app?

src/components/pages/LockScreen.tsx
```diff
@@ -53,10 +53,46 @@ export const LockScreen = (): JSX.Element => {
...
+
+    Promise.all([
+      !biometryEnabled &&
+        webviewIntent.call(
```
Crash--, line 61: nit: I think I'll add a guideline to reject `false && foo()` syntax. if/else clarity for the win.

src/components/pages/LockScreen.tsx
acezard: Due to historical instability with cozy-intent I usually go defensive programming whenever using it

src/components/pages/LockScreen.tsx
acezard: it's still handy when you can't easily use an if statement

src/components/pages/LockScreen.tsx
acezard: No according to React we have to provide a synchronous function returning nothing. Though it looks like a typescript issue in React more than a runtime issue

acezard: I'm converting it to async due to a recent commit from Yannick, it will make the code more readable

PR #514, opened 2022-11-09
refactor: Extract queries from all pages in a file

vdnj:


src/lib/queries.js
```diff
@@ -0,0 +1,46 @@
...
+    as: 'io.cozy.apps/' + slugName,
+    fetchPolicy: CozyClient.fetchPolicies.olderThan(THIRTY_SECONDS),
+    singleDocData: true
+  }
+})
+
+export const queryResultKonnectorsAppPermissions = slugName => ({
+  definition: Q(KONNECTORS_DOCTYPE).getById('io.cozy.konnectors/' + slugName),
+  options: {
+    as: 'io.cozy.konnectors/' + slugName,
```
paultranvan, line 18:
```suggestion
    as: `${KONNECTORS_DOCTYPE}/${slugName}`,
```

src/lib/queries.js
```diff
@@ -0,0 +1,46 @@
...
+    fetchPolicy: CozyClient.fetchPolicies.olderThan(THIRTY_SECONDS),
+    singleDocData: true
+  }
+})
+
+export const queryResultRemoteRequestsAppPermissions = () => ({
+  definition: Q('io.cozy.remote.requests'),
```
paultranvan, line 25: `'io.cozy.remote.requests'` should be added as a const here: https://github.com/cozy/cozy-settings/blob/master/src/doctypes/index.js
```suggestion
  definition: Q(REMOTE_REQUESTS_DOCTYPE),
```
And then,

src/lib/queries.js
```diff
@@ -0,0 +1,46 @@
...
+  options: {
+    as: 'io.cozy.remote.requests',
```
paultranvan, line 27:
```suggestion
    as: REMOTE_REQUESTS_DOCTYPE,
```

src/components/Permissions/DataPermissions/DataPermissions.jsx
```diff
@@ -34,19 +28,20 @@ import ListItem from 'cozy-ui/transpiled/react/MuiCozyTheme/ListItem'
...
-    as: APPS_DOCTYPE,
-    fetchPolicy: CozyClient.fetchPolicies.olderThan(THIRTY_SECONDS)
-  })
-  const queryResultKonnectors = useQuery(Q(KONNECTORS_DOCTYPE), {
-    as: KONNECTORS_DOCTYPE,
-    fetchPolicy: CozyClient.fetchPolicies.olderThan(THIRTY_SECONDS)
-  })
+  const queryResultApps = useQuery(
+    buildAppsQuery().definition,
+    buildAppsQuery().options
```
Ldoppea, line 39: nit: maybe we can avoid duplicate `buildAppsQuery()` call here.
Hopefully those methods are not heavy to compute so it is not really a problem here, but as the method return a combined object it would be great to benefit from it (we can also split the method)

(same comment for all other `buildKonnectorsQuery`, `buildKonnectorsQueryBySlug` etc

Ldoppea (approved):

Ldoppea (approved):

PR #516, opened 2022-11-15
Add uninstall button in AppPermissions

vdnj:


src/components/Permissions/UninstallButton.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react'
+import CircleButton from 'cozy-ui/transpiled/react/CircleButton'
+import Icon from 'cozy-ui/transpiled/react/Icon'
+import TrashIcon from 'cozy-ui/transpiled/react/Icons/Trash'
+import AppLinker, { generateWebLink } from 'cozy-ui/transpiled/react/AppLinker'
+import { useI18n } from 'cozy-ui/transpiled/react/I18n'
+import { useClient } from 'cozy-client'
+
+export const UninstallButton = ({ matchingQueryResultData }) => {
```
Ldoppea, line 9: nit: Here the term `matchingQueryResultData` is too abstract. We don't know what is `matchingQueryResult`.

My understanding is that this is a variable name from the caller.

Maybe you should rename this as `appData`

Then on AppPermissions.jsx you'll have to use
```js
<UninstallButton appData={matchingQueryResult.data} />
```

src/components/Permissions/UninstallButton.jsx
```diff
@@ -0,0 +1,36 @@
...
+  const { subdomain: subDomainType } = client.getInstanceOptions()
+  const appWebRef =
+    matchingQueryResultData &&
+    generateWebLink({
+      cozyUrl: client.getStackClient().uri,
+      slug: 'store',
+      subDomainType,
+      nativePath: `discover/${matchingQueryResultData.slug}/uninstall`
+    })
+  const appData = matchingQueryResultData && matchingQueryResultData
```
Ldoppea, line 21: `matchingQueryResultData && matchingQueryResultData` seems to be useless here. Or maybe I'm missing something?

What'is the difference with `const appData = matchingQueryResultData`?

However, if you apply my previous suggestion, then you would be able to remove this entire line

Ldoppea (approved):

PR #518, opened 2022-11-17
feat: filter apps and konnectors in DataPermissions

vdnj:
Also remove relative imports
Capture d’écran 2022-11-17 à 09 59 28

### DIFF
--- a/src/components/Permissions/DataPermissions/__snapshots__/DataPermissions.spec.jsx.snap
+++ b/src/components/Permissions/DataPermissions/__snapshots__/DataPermissions.spec.jsx.snap
@@ -96,7 +96,7 @@ exports[`DataPermissions should display DataPermissions 1`] = `
                 />
               </div>
               <div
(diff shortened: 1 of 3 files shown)


src/components/Permissions/DataPermissions/__snapshots__/DataPermissions.spec.jsx.snap
```diff
@@ -168,6 +165,9 @@ exports[`DataPermissions should display DataPermissions 1`] = `
               </div>
             </div>
           </li>
+          <hr
+            class="MuiDivider-root MuiDivider-inset"
```
Ldoppea, line 169: Is it intended that the divider also moved with the alphabetical ordering? Shoudln't it still be between `Alan` connector and `Contacts` app?

This seems to be related to `isNotLastItem()` that use an unsorted array for compararison

Ldoppea (approved):

PR #519, opened 2022-11-18
fix: improve sidebar styles

acezard:
### Context

Using the cozy-ui API in list items: https://docs.cozy.io/cozy-ui/react/#/List?id=list-1
In order to maintain stylistic homogeneity

### Implementation

Refactoring existing specificities in cozy-settings and replacing them with generic code

### Testing

Updated existing tests

### Screenshots

image
image

### Thoughts

The result is not pixel perfect to the spec because some differences exist both in cozy-ui (alignment) and cozy-bar (shadow, text alignment)


src/components/Sidebar.styl
```diff
@@ -0,0 +1,6 @@
+@require 'settings/breakpoints'
+
+.Sidebar
+    +gt-tablet()
+        border-right 1px solid var(--dividerColor) // From spec
+        min-width 320px // From spec
```
JF-Cozy, line 6: nit: on peut supprimer le "// From spec" je pense qui pose plus de question que ça n'apporte de réponse je trouve 🤔

JF-Cozy (approved):

src/components/Sidebar.styl
acezard: ça marche

PR #539, opened 2022-12-15
fix: provide instance URL to passphrase function

acezard:
The getDomain() gives a string that is not valid for URL construction.
We decide to leverage the client HoC and inject the correct uri
in the aforementioned functions

### DIFF
--- a/src/containers/Passphrase.jsx
+++ b/src/containers/Passphrase.jsx
@@ -14,6 +14,7 @@ import {
 import { fetchInfos } from 'actions'
 import PassphraseView from 'components/PassphraseView'
 import { useLocation } from 'react-router-dom'
+import { withClient } from 'cozy-client'
 
 const mapStateToProps = state => ({
   fields: state.fields,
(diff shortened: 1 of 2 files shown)


src/containers/Passphrase.jsx
```diff
@@ -86,7 +98,8 @@ const mapDispatchToProps = (dispatch, ownProps) => ({
           currentPassphrase,
           newPassphrase,
           twoFactorCode,
-          twoFactorToken
+          twoFactorToken,
+          ownProps.client.getStackClient().uri
```
Ldoppea, line 102: nit: it may be more "future proof" if we just give the raw `client` and do the `getStackClient().uri` in related methods (but this is very subjective)

Ldoppea (approved):

Ldoppea:
Should we create an issue for other usages of `getStackDomain()` in other files?

acezard:
> Should we create an issue for other usages of `getStackDomain()` in other files?

I'm not sure if the other files need the protocol or just the `//url.url`

src/containers/Passphrase.jsx
acezard: I agree that it's subjective although there's probably an objectively best way. I intended to let the low-level functions be unaware of any cozy-client scope, and handle the URLs I give them. Thus removing complexity from the functions and adding it higher in the container component

PR #552, opened 2023-01-20
fix: Correctly handle navigation when clicking on permission back button

Ldoppea:
With previous implémentation, clicking on the cozy-bar's back button on mobile view would navigate from `#/permissions/slug/:slug` to `#/menu` instead of `#/permissions/slug`

This commit is an attempt to fix the back button behaviour on `AppPermissions.jsx` until we find a more generic way

### DIFF
--- a/src/components/PageTitle.jsx
+++ b/src/components/PageTitle.jsx
@@ -16,10 +16,10 @@ const barTitleStyle = { height: '3rem', display: 'flex', alignItems: 'center' }
 const BarContainer = ({ children }) => {
   return <div style={barTitleStyle}>{children}</div>
 }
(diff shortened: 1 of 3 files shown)


Ldoppea:
What we found:
- `/permissions/:page` can have 2 values `slug` and `data`
  - Both variants are rendered using the same component
  - We can navigate between both variants with a tab component in the page
- `slug` variant can have following children
  - `/permissions/slug/:slug`
  - `/permissions/slug/:slug/:permission`
- `data` variant can have following children
  - `/permissions/data/:permission`


We should find a way to refactor this (not sure how), but for now I will merge this PR as is. The refactor should be done in another PR.

PR #569, opened 2023-02-24
fix: Use displayName instead of Slug

Crash--:
We displayed the technical name (slug) to the end user. We fixed it, by display the right thing to the user

Before: 
Capture d’écran 2023-02-24 à 10 47 17
Capture d’écran 2023-02-24 à 10 47 21

After:

Capture d’écran 2023-02-24 à 10 48 29
Capture d’écran 2023-02-24 à 10 48 32

### DIFF
--- a/src/components/Permissions/AppPermissions/AppPermissions.spec.jsx
+++ b/src/components/Permissions/AppPermissions/AppPermissions.spec.jsx
@@ -172,18 +172,24 @@ describe('AppPermissions', () => {
     expect(queryByTestId('Spinner')).toBeTruthy()
   })
 
-  it('should display slugName when query is not loading', () => {
+  it('should display appName when query is not loading', () => {
     useAppsOrKonnectorsBySlug.mockReturnValue({
       isResultLoading: false,
       hasQueryFailed: false,
-      result: { fetchStatus: 'loaded', data: {} }
+      result: {
+        fetchStatus: 'loaded',
+        data: {
+          name: 'Drive',
+          name_prefix: 'Coz'
+        }
+      }
     })
     const { queryByText } = render(
       <BreakpointsProvider>
         <AppPermissions />
       </BreakpointsProvider>
     )
-    expect(queryByText('DRIVE')).toBeInTheDocument()
+    expect(queryByText('Coz Drive'.toUpperCase())).toBeInTheDocument()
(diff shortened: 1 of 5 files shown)


src/components/Permissions/hooks/useAppsOrKonnectorsBySlug.tsx
```diff
@@ -19,7 +19,7 @@ interface useAppsOrKonnectorsBySlugReturns {
 }
 
 const useAppsOrKonnectorsBySlug = (
-  slugName: string
+  slugName: string | undefined
```
cballevre, line 22: why the slugName can be undefined ?

src/components/Permissions/AppPermissions/AppPermissions.spec.jsx
```diff
@@ -172,18 +172,24 @@ describe('AppPermissions', () => {
...
+        }
+      }
     })
     const { queryByText } = render(
       <BreakpointsProvider>
         <AppPermissions />
       </BreakpointsProvider>
     )
-    expect(queryByText('DRIVE')).toBeInTheDocument()
+    expect(queryByText('Coz Drive'.toUpperCase())).toBeInTheDocument()
```
cballevre, line 192: nit: maybe we can use toBeTruthy instead toBeInTheDocument ?

cballevre (approved):

src/components/Permissions/AppPermissions/AppPermissions.spec.jsx
Crash--: > nit: maybe we can use toBeTruthy instead toBeInTheDocument ?

There is a lot of things that are not OK within this test file. `toBeIntheDocument` is the latest thing I'll change ^^'. 

For instance, tests are not isolated. This one need the first to be run otherwise it fails. Mocks returned values are not OK, we don't return only data with an empty object... Lot of console.error...

src/components/Permissions/hooks/useAppsOrKonnectorsBySlug.tsx
Crash--: In fact it's not, but I had a TS issue on the caller. I'll revert this change since we don't care ATM

PR #572, opened 2023-03-21
Feat/change default redirection

zatteo:
```
### ✨ Features

* Change your default app in profile section
```

Related : 
- https://github.com/cozy/cozy-libs/pull/2069
- https://github.com/cozy/cozy-flagship-app/pull/682

### DIFF
--- a/src/components/DefaultRedirectionSection.jsx
+++ b/src/components/DefaultRedirectionSection.jsx
@@ -9,6 +9,8 @@ import {
 
 import Select from 'components/Select'
 import { buildAppsQuery } from 'lib/queries'
+import { useWebviewIntent } from 'cozy-intent'
+import { isFlagshipApp } from 'cozy-device-helper'
 
 const EXCLUDED_SLUGS = ['settings', 'store', 'home']
 
@@ -48,6 +50,7 @@ const formatDefaultRedirection = slug => `${slug}/`
 
 const DefaultRedirectionSection = props => {
   const { t } = useI18n()
+  const webviewIntent = useWebviewIntent()
   const appsQuery = buildAppsQuery()
   const appsResult = useQuery(appsQuery.definition, appsQuery.options)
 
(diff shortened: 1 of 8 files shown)


src/components/DefaultRedirectionSection.jsx
```diff
@@ -0,0 +1,100 @@
...
+
+const formatOptions = (appsResult, t) => {
+  if (!hasQueryBeenLoaded(appsResult)) return []
+
+  return appsResult.data
+    .filter(({ slug }) => !EXCLUDED_SLUGS.includes(slug))
+    .map(({ slug, name }, index) => {
+      if (slug === 'home') {
+        return {
+          index: -1, // to set 'home' before everything else when sorting
```
Ldoppea, line 24: Two questions:
- Should we move this logic in the `.sort()` instead? Using a .map to do a sort is a bit odd
- Don't we sort other apps alphabetically?

Ldoppea: After reading a bit more, I understand that we do the map in order to rename the `home` option, so it is more optimal to also do the index thing here.
But the question still remain about the alpabetical sort

src/components/DefaultRedirectionSection.jsx
```diff
@@ -0,0 +1,100 @@
...
+            value: app.slug,
+            label: app.name
+          }
+        })}
+        fieldProps={fieldProps}
+        value={selectedSlug}
+        onChange={sel => {
+          const newDefaultRedirection = formatDefaultRedirection(sel.value)
+          onChange(fieldName, newDefaultRedirection)
+          webviewIntent.call('setDefaultRedirection', newDefaultRedirection)
```
Ldoppea, line 93: Should we condition this call to `isFlagshipApp()`?

zatteo: Yes I think

src/components/DefaultRedirectionSection.jsx
zatteo: When I checked they are returned alphabetically sorted by the stack so I was counting on it.

Ldoppea: Ok.

@Crash-- do you know if this is a "lucky" effect or if we can consider this to be the norm?

However this is a detail that should not block the PR so I will approve.

Ldoppea (approved):

PR #576, opened 2023-04-04
feat(DeleteAccount): Add a modal to inform about email confirmation

cballevre:
```
### ✨ Features

* Add a modal to inform about email confirmation after delete account success
```


src/components/DeleteAccount.jsx
```diff
@@ -8,12 +8,14 @@ import { useI18n } from 'cozy-ui/transpiled/react/I18n'
 import ConfirmModal from 'components/DeleteAccount/ConfirmModal'
 import FormModal from 'components/DeleteAccount/FormModal'
 import { useCanAuthWith } from 'hooks/useCanAuthWith'
+import EmailConfirmationModal from './DeleteAccount/EmailConfirmationModal'
```
Crash--, line 11: alias

src/components/DeleteAccount.jsx
```diff
@@ -8,12 +8,14 @@ import { useI18n } from 'cozy-ui/transpiled/react/I18n'
...
 
 const CONFIRMING = 'confirming'
 const IDLE = 'idle'
 const REQUESTING = 'requesting'
+const EMAIL_CONFIRMATION = 'email_confirmation'
 
-const DeleteAccount = () => {
+const DeleteAccount = ({ email }) => {
```
Crash--, line 18: nit: propTypes or jsdoc 😗

src/components/ProfileView.jsx
```diff
@@ -154,7 +154,7 @@ const ProfileView = props => {
             </div>
           </Stack>
           <div className="u-mt-2">
-            <DeleteAccount />
+            <DeleteAccount email={instance?.data?.attributes?.email} />
```
Crash--, line 157: You should not check if instance.data.attributes are there. You're in a `!isFetching` condition. This `isFetching` is set when the `fetchInfos` is launched. After that, is set to false when the request is done. 

Also it's weird to be able to pass `undefined` to DeleteAccount, since we'll display the modal without any email.

Crash-- (approved):

PR #579, opened 2023-04-06
fix: Instance can be null during first renders leading to error

zatteo:


zatteo:
Update : https://github.com/cozy/cozy-settings/pull/576#discussion_r1158075984 

Investigating more...

cballevre (approved):
Thx 🙏

PR #587, opened 2023-05-23
fix: Update instance data after 2FA activation

cballevre:
This patch is the improvement around after the review of https://github.com/cozy/cozy-settings/pull/586. Unfortunately, the PR was automatically merged before the new changes were pushed. Therefore I revert the commit from the previous PR before reapply new changes.

### DIFF
--- a/src/reducers/fields.js
+++ b/src/reducers/fields.js
@@ -8,37 +8,8 @@ import {
   RESET_INFO_FIELD
 } from 'actions'
 
-import {
-  CHECK_TWO_FACTOR_CODE_SUCCESS,
-  DESACTIVATE_2FA_SUCCESS,
-  AUTH_MODE
-} from 'actions/twoFactor'
-
-const composeReducers = (...fns) => {
-  return (state, action) => {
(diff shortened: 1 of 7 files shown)


src/reducers/fields.js
```diff
@@ -8,37 +8,8 @@ import {
...
-const authModeValue = (state = AUTH_MODE.BASIC, action) => {
-  switch (action.type) {
-    case CHECK_TWO_FACTOR_CODE_SUCCESS:
-      return AUTH_MODE.TWO_FA_MAIL
-    case DESACTIVATE_2FA_SUCCESS:
-      return AUTH_MODE.BASIC
-    default:
-      return state
-  }
-}
```
zatteo, line 38: What was the purpose of these functions?

cballevre: Before the refactor of the 2FA part, we managed all requests and their status with redux. We also determined from this status whether 2FA had been activated. Now, we determine directly in the 2FA component whether it is activated or not. I decided to remove this part because it was no longer useful as we no longer have a state in redux

zatteo (approved):

PR #588, opened 2023-06-20
Improve password management

cballevre:
```
### ✨ Features

* Allow password modification for all instances where it is defined
* Disable password confirmation when deleting an account if no password is defined

### 🐛 Bug Fixes

* Correctly handling errors from cozy-client

### 🔧 Tech

* Add CozyDevTools from cozy-client in development mode
```

**Related PRs:**
- https://github.com/cozy/cozy-stack/pull/4020
- https://github.com/cozy/cozy-client/pull/1361


src/components/PasswordSection.spec.jsx
```diff
@@ -23,7 +26,7 @@ describe('PasswordSection', () => {
     expect(root.getByText('Update my password')).toBeTruthy()
   })
 
-  it('should render a button if we can auth with password (defaults to true)', () => {
+  it.skip('should render a button if we can auth with password (defaults to true)', () => {
```
Crash--, line 29: Comment ça ? :D

Crash--:
@Ldoppea can you take a look at the `defined password` stuff? :)

src/actions/passphrase.js
```diff
@@ -35,7 +35,10 @@ const defaultErrorAction = {
 export const getErrorDetails = error => {
   const vaultError = error && error.response && error.response.error
   const stackError =
-    error && error.errors && error.errors[0] && error.errors[0].detail
+    error &&
+    error.reason.errors &&
+    error.reason.errors[0] &&
+    error.reason.errors[0].detail
```
Ldoppea, line 41: nit: Seems like this project now supports the `?.` operator, so `const stackError = error?.reason?.error?.[0].detail` should work

src/actions/passphrase.js
```diff
@@ -167,7 +170,7 @@ export const updatePassphrase2FASecond = (
       dispatch({ type: UPDATE_PASSPHRASE_2FA_2_SUCCESS })
       dispatch({ type: RESET_PASSPHRASE_FIELD })
     } catch (error) {
-      const errors = error.errors || []
+      const errors = (error.reason && error.reason.errors) || []
```
Ldoppea, line 173: Same: `const errors = error.reason?.errors || []`

src/components/App.jsx
```diff
@@ -93,6 +94,7 @@ export const App = () => {
         </Routes>
       </Main>
       <Sprite />
+      {process.env.NODE_ENV !== 'production' ? <CozyDevtools /> : null}
```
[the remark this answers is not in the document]
cballevre: J'ai mis à jour pour que cela soit dispo en prod. J'ai fait une [PR sur drive](https://github.com/cozy/cozy-drive/pull/2926) pour avoir le même comportement

src/components/PasswordSection.spec.jsx
cballevre: Reliquat de développement en cours ce test disparait dans la nouvelle version

src/actions/passphrase.js
cballevre: Same 👌

src/actions/passphrase.js
cballevre: Yes way nicer 👌

src/actions/import.js
```diff
@@ -20,7 +20,7 @@ export const precheckImport = (client, url) => {
       dispatch({ type: PRECHECK_IMPORT_SUCCESS })
     } catch (e) {
       let error = 'server_error'
-      switch (e.errors && e.errors[0] && +e.errors[0].status) {
+      switch (e.status) {
```
Ldoppea, line 23: You are loosing the int conversion done by the `+` operator. Did you check that the switch case still work without it?

src/actions/index.js
```diff
@@ -103,10 +103,10 @@ export const fetchStorageData = client => {
         offersLink = `${managerUrl}/cozy/instances/${uuid}/premium`
       }
     } catch (e) {
-      if (e.error && e.error !== 'Not Found') {
+      if (e.reason.error && e.reason.error !== 'Not Found') {
         // eslint-disable-next-line no-console
         console.warn(e)
-      } else if (!e.error) {
+      } else if (!e.reason.error) {
```
Ldoppea, line 109: I don't understand that condition.

As we do exactly the same thing on both "if" and "else if" block, can this be simplified by this?
```js
if (e.reason.error !== 'Not Found') {
  console.warn(e)
}
```

Also do you know the reason why we don't log anything in case of 'Not Found'?

src/actions/index.js
```diff
@@ -135,16 +135,15 @@ const tryUpdate = async (client, instance, { retries = 0 }) => {
   try {
     return await cozyFetch(client, 'PUT', '/settings/instance', instance)
   } catch (error) {
-    const isConflictError =
-      error.errors && error.errors[0] && error.errors[0].status === '409'
+    const isConflictError = error.status === 409
```
Ldoppea, line 138: Does the fact that you know use `error.status` instead of `error.errors[0].status` means that this code did not work before?

Ldoppea:
> The property password_defined of instance is not reactive with realtime because it is the stack that defines it. I manage the state locally via a provider to update user interface consistently

I'm not sure why we use a hook with a state as the methods you are using are called inside of a workflow and doesn't seems to be used in the UI. What is the advantage compared to a single "just in time" call?

src/components/DeleteAccount.jsx
```diff
@@ -32,10 +32,10 @@ const DeleteAccount = ({ email }) => {
     setStatus(EMAIL_CONFIRMATION)
   }
 
-  const { canAuthWithOIDC } = useCanAuthWith()
+  const withoutPassword = useWithoutPassword()
 
   const onClick = () => {
-    if (canAuthWithOIDC) {
+    if (withoutPassword) {
```
Ldoppea, line 38: What is the expected behavior here?

If an OIDC user defines a password, do we expect the user to enter their password on deleting account? Or is the OIDC login enough?

src/components/PasswordSection.jsx
```diff
@@ -1,24 +1,17 @@
...
-  // If we have no capabilities, consider that we can login to the cozy with password
-  const canAuthWithPassword = get(
-    client,
-    'capabilities.can_auth_with_password',
-    true
-  )
+  const withoutPassword = useWithoutPassword()
 
-  return canAuthWithPassword ? (
+  return !withoutPassword ? (
```
Ldoppea, line 14: Does this mean that an OIDC user can never set a password? I think i'm missing something as the rest of the code is constructed as if it was possible.

Ldoppea:
Nit: `withoutPassword` make the code a bit difficult to reason about. It produces double negations like in PasswordSection (`return !withoutPassword ? ... : ...` )?

I would suggest `isPasswordless` but I already used it for OIDC/Magiclink instances no matter if they defined a password or not ([cf](https://github.com/cozy/cozy-flagship-app/blob/db0f237cb6719674a0311eda8da8a5d73953b58c/src/app/domain/authentication/models/User.ts#L13))

We may need to homogenize our codebases. On your side we have:
- hasDefinedPassword -> check in couchdb if password_defined is true
- withoutPassword -> hasDefinedPassword + check if OIDC/magiclink

On [cozy-flagship-app](https://github.com/cozy/cozy-flagship-app/blob/db0f237cb6719674a0311eda8da8a5d73953b58c/src/app/domain/authentication/models/User.ts#L13) we have:
- hasDefinedPassword -> check in couchdb if password_defined is true
 + check if is not OIDC/magiclink
- isPasswordLessCozy -> check if OIDC/magiclink

Maybe we should make a visio on this to define a common nomenclature?

src/components/PasswordSection.jsx
Ldoppea: Answer: the settings's option "edit password" is hidden for a passwordless cozy, but the user will be able to set a password when activating the 2FA (then the password IS the 2FA)

src/actions/import.js
cballevre: Yes e.status return a directly that's why I made this simplification

src/actions/index.js
cballevre: Me either but we could simplify it by returning `console.warn(e)` directly to see if there is a 404 problem 🤔

src/actions/index.js
cballevre: This code worked before I modified cozyFetch to use cozy-client under the hood. The code would look like ```
error?.reason?.errors?.[0]?.status
```
I chose to use directly status from the query to simply the code as it return the same result

**Comment by cballevre** [comment]
After discussions with @Ldoppea, we decided to harmonise our code base with the `hasPassword` name. The business logic is in cozy-client in the [user model](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/models/user.js) to facilitate reuse.

**Comment by cballevre** [comment]
@Crash-- or @Ldoppea could I have a new review of the changes I've made? pls 🙏

File: src/hooks/useHasPassword.js [hunk 10]
```diff
@@ -0,0 +1,42 @@
...
+
+  const handlePassphraseCreated = settings => {
+    if (settings._id === 'io.cozy.settings.passphrase') {
+      setHasPassword(true)
+    }
+  }
+
+  useEffect(() => {
+    client.plugins.realtime.subscribe(
+      'created',
```
Ldoppea, line 21: Is this always `created` here? My comprehension was that `io.cozy.settings.passphrase` was always existing but it was just random for OIDC and MagicLink users, so defining a password would consist of replacing the random unknown password by a known one. With that in mind, shouldn't it be `updated`?

Ldoppea (approved):

src/actions/index.js
Ldoppea: I think we should simplify it 👍

PR #604, opened 2023-07-24
fix: Change subscription flag for office to write instead enabled

cballevre:


zatteo:
To update on the wiki ?

cballevre:
This flag is already documented in the wiki its only my mistake about how to differentiate between freemium and premium for office

PR #629, opened 2023-09-05
refactor: Remove redux into profile

cballevre:
This PR removes the need to use redux to retrieve or modify `io.cozy.settings.instance`. It now uses cozy-client with the save method to modify it. I've created a wrapper to this method to handle its state, inspired by the `useMutation` from react-query ([doc](https://tanstack.com/query/latest/docs/react/guides/mutations)).


src/components/Profile/TrackingSection.jsx
```diff
@@ -1,25 +1,46 @@
...
     <Input
       name="tracking"
       type="checkbox"
       title={t('ProfileView.tracking.title')}
       label={t('ProfileView.tracking.label', {
-        version:
-          instance && instance.data.attributes.tos
-            ? `-${instance.data.attributes.tos}`
-            : '-201711'
+        version: instance?.tos ? `-${instance.tos}` : '-201711'
```
zatteo, line 37: No more data.attributes ?

zatteo:
Maybe your `useMutation` should added in cozy-client? @paultranvan What do you think?

zatteo (approved):

src/components/Profile/TrackingSection.jsx
cballevre: To work in real time, attributes are spread at the root of the object ([code](https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/SettingsCollection.js#L13-L19)). Normally this should be the norm everywhere from what I understand, but migration is complex ([discussion about](https://github.com/cozy/cozy-client/issues/1118))

zatteo: Thanks for the explanation!

src/components/2FA/ActivationConfirmation.jsx
```diff
@@ -5,15 +5,19 @@ import { Media, Img, Bd } from 'cozy-ui/transpiled/react/deprecated/Media'
 
 import ReactMarkdownWrapper from 'components/ReactMarkdownWrapper'
 
-export const ActivationConfirmation = ({ images, error }) => {
+const twoFaModalBanner = require('assets/images/double_authent_prez_banner.svg')
+const twoFaModalProtect = require('assets/images/protect_data_point.svg')
+const twoFaModalSecu = require('assets/images/niv_secu_point.svg')
```
Crash--, line 10: you can use import in order to be coherent (and like that we'll not mix esm & commonjs)

src/components/Profile/LanguageSection.jsx
```diff
@@ -2,45 +2,73 @@ import React from 'react'
...
 const LANG_OPTIONS = ['en', 'fr', 'es']
 
-const LanguageSection = props => {
+const LanguageSection = () => {
   const { t } = useI18n()
-  const { fields, onChange } = props
+  const dispatch = useDispatch()
+  const { mutate, isLoading, isSuccess, error } = useMutation({
+    onSuccess: data => {
+      dispatch({ type: SET_LANG, lang: data.locale })
```
Crash--, line 21: next pr, remove the need to dispatch something? :D

src/containers/Profile.jsx
```diff
@@ -2,28 +2,20 @@ import { connect } from 'react-redux'
...
-  fields: state.fields,
   passphrase: state.passphrase,
-  instance: state.instance,
   exportData: state.exportData,
   importData: state.importData,
   isFetching: state.ui.isFetching
 })
 
 const mapDispatchToProps = (dispatch, ownProps) => ({
-  fetchInfos: () => dispatch(fetchInfos(ownProps.client)),
```
Crash--, line 23: does the `fetchInfos` is still needed somewhere? Can we delete it?

src/actions/index.js
```diff
@@ -1,7 +1,6 @@
 /* eslint-disable promise/always-return */
 // eslint-disable-next-line no-redeclare
 
-import emailHelper from 'lib/emailHelper'
```
Crash--, line 4: can we delete this lib/emailHelper file?

Crash-- (approved):
Nice work! As said on zoom, I'm pretty excited about the useMutation hook. It just need a few adjustments to have a better match with the useQuery API but it'll be a great addition to cozy-client. 

Also, I really appreciate the work you do on this app progressively! \o/

src/containers/Profile.jsx
cballevre: It still used into intent but I will something to tackle in my next PR. I'd like to see if they can be removed as it was used for Claudy button and it was removed from the cozy-bar

src/components/Profile/LanguageSection.jsx
cballevre: Yes for sure

src/components/2FA/ActivationConfirmation.jsx
cballevre: It's done 🙂

src/actions/index.js
cballevre: No we are still using it into the EmailModal to check the email format

PR #634, opened 2023-09-19
WIP: rspack build

Crash--:
- [x] working build (yarn run rspack build) 
- [ ] public asset (favico etc) 
- [ ] ejs support template 
- [ ] chunks (app / vendors) 
- [ ] mode (prod / dev) 
- [ ] postcss (really needed ?)


cballevre:
We have decided to use Rsbuild rather than Rspack directly (https://github.com/cozy/cozy-settings/pull/727)

PR #650, opened 2024-01-04
[VO-167] feat(Subscription): Show max for CoachCO2 if defined

cballevre:
```
### ✨ Features

* Show max for CoachCO2 if defined in Subscription page
```

### DIFF
--- a/src/components/Subscription/helper.js
+++ b/src/components/Subscription/helper.js
@@ -34,6 +34,12 @@ export function computeFlagLabel(name, t, lang) {
     return `${beginning}${t(`Subscription.included.numeric_flag_unlimited`)}`
   }
 
+  if (name === 'coachco2.max-days-to-capture') {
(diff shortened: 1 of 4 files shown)


src/components/Subscription/helper.js
```diff
@@ -34,6 +34,12 @@ export function computeFlagLabel(name, t, lang) {
     return `${beginning}${t(`Subscription.included.numeric_flag_unlimited`)}`
   }
 
+  if (name.includes('max-days')) {
```
zatteo, line 37: nit: Why include and not a strict equality with flag name ?

zatteo (approved):

src/components/Subscription/helper.js
cballevre: That's a good point. I was going for a more generic case, but it doesn't make sense

PR #654, opened 2024-01-31
[VO-327] feat: Add sentry

cballevre:
```
### 🔧 Tech

* Add sentry to monitor crash event
```


zatteo:
I did not know about Sentry x React Router. Interesting. Curious to see what we will get and if it is useful.

zatteo (approved):

PR #658, opened 2024-02-16
[VO-392] fix(LockScreen): Disable ellipsis

cballevre:
```
### 🐛 Bug Fixes

* Disable ellipsis on the LockScreen
```

### DIFF
--- a/src/cozy-ui.d.ts
+++ b/src/cozy-ui.d.ts
@@ -15,7 +15,7 @@ declare module 'cozy-ui/transpiled/react/Divider' {
 
 declare module 'cozy-ui/transpiled/react/ListItem' {
   import ListItem from '@material-ui/core/ListItem'
-  export default ListItem
+  export default ListItem as ListItem & { ellipsis: boolean }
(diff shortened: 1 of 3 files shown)


Ldoppea (approved):

src/cozy-ui.d.ts
```diff
@@ -15,7 +15,7 @@ declare module 'cozy-ui/transpiled/react/Divider' {
 
 declare module 'cozy-ui/transpiled/react/ListItem' {
   import ListItem from '@material-ui/core/ListItem'
-  export default ListItem
+  export default ListItem as ListItem & { ellipsis: boolean }
```
Ldoppea, line 18: Wondering if we should keep track of all types we should fix on cozy-ui and cozy-client somewhere?

cballevre: Very good question but I don’t have clear view where to put it. Maybe should we focus on adding jsdoc on cozy-ui dans generate type there 🤔

PR #715, opened 2024-08-28
[VO-1038] chore: Migrate to rspack

cballevre:
This is preparatory work to see what the new configuration would be if we migrated to Rspack. This work had already been started on this PR: https://github.com/cozy/cozy-settings/pull/634


cballevre:
We have decided to use Rsbuild rather than Rspack directly (https://github.com/cozy/cozy-settings/pull/727)

PR #679, opened 2024-03-29
[VO-591] chore: Make continuous deployement

cballevre:
This commit add semantic-release to help us publish new version on every branch merge into master.

We add a commit with the modification of the version in the `package.json` and the `manifest.webapp`. This commit has the tag of the new version associated with it. After we still use cozy-app-publish to push the build to the registry and downcloud.

```
### 🔧 Tech

* Make continuous deployement
```

PR #717, opened 2024-10-16
chore: Add CI with Github Actions

Merkur39:
Remove Travis and continuous deployment implemented here #679 (to be iso with other apps).

```
### 🔧 Tech

* Remove Travis
* Remove continuous deployment
```


cballevre (approved):
LGTM 👍  We just need to found a place to keep a track of the continus deployement experiment

PR #733, opened 2025-04-14
Add Content from Layout and rework sidebar for desktop

JF-Cozy:


src/components/SidebarDesktop.jsx
```diff
@@ -0,0 +1,205 @@
...
+  const { isLoaded, instance, diskUsage } = useInstanceInfo()
+  const { canOpenPremiumLink, premiumLink } = usePremium()
+
+  const NavLink = genNavLinkForV6(RouterLink)
+  const percent = isLoaded
+    ? makeDiskInfos(diskUsage.data.used, diskUsage.data.quota).percentUsage
+    : ''
+
+  return (
+    <Sidebar style={{ width: 360 }}>
```
zatteo, line 50: Can we do direct styling like this?

JF-Cozy: idealy we need to modify Sidebar in cozy-ui to be able to pass classes, but there's no need to do that except in Settings, so this is a quick answer

PR #735, opened 2025-04-15
Force value for sidebarHeight

JF-Cozy:
The variable `--sidebarHeight` is used to calculate the heights of the layout elements. However, during the build with rsbuild, the value `0px` or `0rem` is replaced by `0`, which causes issues in cases like `calc(100vh - var(--sidebarHeight))`. While waiting to fix the problem on the rsbuild side, here is a small trick to prevent the layout from breaking.


zatteo:
Can you add in the first commit message the very clear PR description?

zatteo (approved):

PR #741, opened 2025-04-22
feat(UI): Add copy icon, 2FA component flags and delete flow confirma…

lenhanphung:
- Add copy icon to input field components
- Implement flags into 2FA component
- Add button for changing phone number in phone number component
- Add delete confirmation popup to account deletion flow

### DIFF
--- a/src/components/Field.jsx
+++ b/src/components/Field.jsx
@@ -1,12 +1,17 @@
 import classNames from 'classnames'
-import React from 'react'
+import React, { useState, useEffect, useRef } from 'react'
 
 import Icon from 'cozy-ui/transpiled/react/Icon'
+import IconButton from 'cozy-ui/transpiled/react/IconButton'
 import CheckIcon from 'cozy-ui/transpiled/react/Icons/Check'
+import CopyIcon from 'cozy-ui/transpiled/react/Icons/Copy'
 import Spinner from 'cozy-ui/transpiled/react/Spinner'
 import Typography from 'cozy-ui/transpiled/react/Typography'
 import { Media, Bd, Img } from 'cozy-ui/transpiled/react/deprecated/Media'
+import { useAlert } from 'cozy-ui/transpiled/react/providers/Alert'
+import { useCozyTheme } from 'cozy-ui/transpiled/react/providers/CozyTheme'
 import { useI18n } from 'cozy-ui/transpiled/react/providers/I18n'
+import { useTheme } from 'cozy-ui/transpiled/react/styles'
 
 import styles from '@/styles/fields.styl'
 
@@ -14,13 +19,66 @@ import ReactMarkdownWrapper from '@/components/ReactMarkdownWrapper'
 
 const absoluteStyle = {
   position: 'absolute',
-  right: '1rem',
-  top: '1rem',
-  bottom: 0,
-  margin: 'auto'
+  top: '0.5rem',
+  right: 0
 }
 
-const StatusIcon = ({ submitting, saved, className, absolute }) => {
+const StatusIcon = ({
+  submitting,
+  saved,
+  className,
+  absolute,
+  copyable,
+  value,
+  onCopy
+}) => {
+  const { t } = useI18n()
+  const { isLight } = useCozyTheme()
+  const theme = useTheme()
+  const { showAlert } = useAlert()
+  const [showCopyIcon, setShowCopyIcon] = useState(true)
+  const timerRef = useRef(null)
+
+  useEffect(() => {
+    if (saved && copyable) {
+      setShowCopyIcon(false)
+
+      if (timerRef.current) {
+        clearTimeout(timerRef.current)
+      }
+
+      timerRef.current = setTimeout(() => {
+        setShowCopyIcon(true)
+      }, 2000)
+    }
+
+    return () => {
+      if (timerRef.current) {
+        clearTimeout(timerRef.current)
+      }
+    }
+  }, [saved, copyable])
+
+  const handleCopy = async () => {
+    if (value && onCopy) {
+      onCopy(value)
+    } else if (value) {
+      try {
+        await navigator.clipboard.writeText(value)
+        showAlert({
+          message: t('Input.copied', 'Copied to clipboard'),
+          type: 'success'
+        })
+      } catch (err) {
+        console.error('Could not copy text: ', err)
+        showAlert({
+          message: t('Input.copy_failed', 'Failed to copy to clipboard'),
+          type: 'error'
+        })
+      }
+    }
+  }
+
   return (
     <span className={className} style={absolute && absoluteStyle}>
       {submitting && (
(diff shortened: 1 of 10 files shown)


src/components/Profile/ProfileView.jsx
```diff
@@ -43,6 +44,8 @@ const ProfileView = ({
     instanceQuery.options
   )
 
+  const isTwoFAEnabled = flag('settings.2fa.enabled') !== false
```
zatteo, line 47: `flag('settings.2fa.enabled')` can return a boolean, so it can return `true` or `false` or `undefined` or `null` so no need to add `!== false` normally

src/components/Profile/PhoneNumberSection.jsx
```diff
@@ -53,18 +55,27 @@ const PhoneNumberSection = () => {
...
+        saved={mutationStatus === 'loaded'}
+        errors={errors}
+        copyable={true}
+      />
+      <Button
+        variant="secondary"
+        size="medium"
+        label={t('ProfileView.phone_number.change_button')}
+        href="https://sign-up.twake.app/"
+      />
```
zatteo, line 77: What does it mean? That you can change anymore your phone number from here and you need to go to `https://sign-up.twake.app/` to do it?

src/components/Profile/DeleteSection.jsx
```diff
@@ -1,16 +1,101 @@
...
+      navigate('delete')
+    }
+  }
+
+  const handleClose = () => {
+    setCurrentPopup(0)
+  }
+
+  const popupData = [
+    {},
```
zatteo, line 35: Is it necessary ? Because you already have `if (currentPopup === 0) return null` ?

src/components/Field.jsx
```diff
@@ -1,26 +1,95 @@
...
+
+  const handleCopy = () => {
+    if (value && onCopy) {
+      onCopy(value)
+    } else if (value) {
+      if (navigator && navigator.clipboard && navigator.clipboard.writeText) {
+        navigator.clipboard
+          .writeText(value)
+          .then(() => {
+            Alerter.success(t('Input.copied', 'Copied to clipboard'))
```
zatteo, line 69: Could you use `useAlert` from `import { useAlert } from 'cozy-ui/transpiled/react/providers/Alert' instead of deprecated Alerter ?

See documentation here https://docs.cozy.io/cozy-ui/react/#!/Providers/AlertProvider

src/components/Profile/AvatarSection.jsx
```diff
@@ -222,38 +206,6 @@ const AvatarSection = () => {
           </div>
         </div>
       </div>
-
-      {isConfirmDeleteDialogOpen && (
-        <ConfirmDialog
-          open={isConfirmDeleteDialogOpen}
-          onClose={closeDeleteConfirmation}
-          title={t('AvatarSection.deleteConfirm.title', 'Confirm Deletion')}
```
zatteo, line 230: Do not forget to remove the translation if they are not used anymore.

src/components/Field.jsx
```diff
@@ -1,26 +1,95 @@
...
+        document.body.appendChild(textArea)
+        textArea.focus()
+        textArea.select()
+        try {
+          document.execCommand('copy')
+          Alerter.success(t('Input.copied', 'Copied to clipboard'))
+        } catch (err) {
+          console.error('Fallback: Could not copy text: ', err)
+        }
+        document.body.removeChild(textArea)
```
zatteo, line 88: Why this `textArea` ? I think supporting copy with `navigator.clipboard.writeText` is enough.

src/components/Field.jsx
```diff
@@ -29,11 +98,39 @@ const StatusIcon = ({ submitting, saved, className, absolute }) => {
...
-        />
+      {saved && !showCopyIcon && (
+        <IconButton
+          size="small"
+          className="u-mr-half"
+          style={{
+            color: isLight
+              ? theme.palette.text.secondary
+              : theme.palette.text.primary,
+            zIndex: 1
```
zatteo, line 109: Is the `zIndex` necessary ?

src/components/Profile/PhoneNumberSection.jsx
lenhanphung: https://sign-up.twake.app/ is temporary link for now? I also confirm with Bob. lets talk about this in daily

src/components/Profile/DeleteSection.jsx
lenhanphung: I think it's still necessary:

1. The `popupData` array serves as a data structure that stores all the popup information (icon, title, description) in one place, making it easier to maintain and modify.

2. While `if (currentPopup === 0) return null` only checks if any popup should be shown, the `popupData` array is essential for actually rendering the correct popup content when `currentPopup` is not 0.

3. Without `popupData`, we would need to write multiple conditional statements or switch cases to handle each popup case, which would make the code more verbose and harder to maintain.

For example, without `popupData`, the code would look like this:
```javascript
if (currentPopup === 1) {
  return <ConfirmDialog icon={ChatIcon} title={t('...')} description={t('...')} />
} else if (currentPopup === 2) {
  return <ConfirmDialog icon={ChatIcon} title={t('...')} description={t('...')} />
} 
// and so on for other popups...
```

Therefore, I believe keeping the `popupData` array is a better approach as it:
- Makes the code more maintainable
- Reduces duplication
- Makes it easier to add or modify popup content
- Follows the DRY (Don't Repeat Yourself) principle

Would you like me to make any specific improvements to the current implementation?

src/components/Profile/ProfileView.jsx
lenhanphung: Yes. Updated

src/components/Field.jsx
lenhanphung: yes. updated

src/components/Profile/AvatarSection.jsx
lenhanphung: yes. updated

src/components/Field.jsx
lenhanphung: updated. also update for avatar section

src/components/Field.jsx
lenhanphung: updated, but I think keep this as fall back if in anycase navigator.clipboard.writeText doesn't work.

src/components/Field.jsx
lenhanphung: yes. I see it under input if doesnt set zIndex

src/components/Profile/DeleteSection.jsx
zatteo: I like the `popupData` approach, just wondering about the necessity of the first special case `{}`

src/components/Field.jsx
zatteo: I really think writeText is enough. It is supported by all browser for 4 or 5 years and I would prefer reducing code complexity here if it does not increase a lot browser support.
image

src/components/Field.jsx
zatteo: Awesome 👍

src/components/Profile/PhoneNumberSection.jsx
zatteo: Discussed in daily, temporary link to show what will be possible

src/components/Profile/DeleteSection.jsx
lenhanphung: you right. remove it

src/components/Field.jsx
lenhanphung: agree

PR #765, opened 2025-09-30
New email only method to delete account

zatteo:
https://github.com/user-attachments/assets/6f7df93e-d7d4-4ed6-af2f-a5d4504ee531


src/components/DeleteAccount/DeleteAccount.jsx
```diff
@@ -42,11 +43,23 @@ const DeleteAccount = () => {
...
+    })
   }
 
-  const onRequested = () => {
-    setStatus(EMAIL_CONFIRMATION)
+  const onRequested = ({ byEmailOnly }) => {
+    if (byEmailOnly) {
+      showAlert({
+        message:
+          'La demande a été envoyée et sera traitée rapidement par nos équipes.',
```
rezk2ll, line 56: this should be translated i guess.

edit: i found the string should be byEmailOnly.success

PR #783, opened 2025-10-20
Feat/add services imports

KillianCourvoisier:
This PR introduces the full integration of the new **Data Imports** feature into Cozy Settings.

A new entry has been added in the “Data” menu, with three dedicated pages:

- `/imports`
  - Lets the user enable or disable the import tools.

- `/imports/run`
  - Allows the selection of a provider (only Nextcloud for now — the others are displayed as placeholders).
  - Supports account selection, creation and deletion through a modal. Multiple accounts from different Nextcloud instances can be connected.
  - Lets the user choose the exact remote path (root, directory, or file) and preview the first 10 items to verify the content before launching an import.
  - Displays a real-time progress bar (processed files, total files, and current file), dynamically updating depending on the folder being scanned.
  - Shows the list of errors encountered during the import, with filename and status code, to help understand why specific items were skipped.
  - Allows the user to manually stop the import at any time.

- `/imports/history`
  - Currently empty and will be implemented later.

The entire feature is wrapped behind a Cozy feature flag (`settings.imports`).  
If the flag is off, the menu entry and routes are fully hidden.

For now, only Nextcloud accounts are supported. Additional providers such as Google Drive and Dropbox will be added in future updates.


JF-Cozy:
@KillianCourvoisier don't hesitate to ping us again when it's ready, I think github doesn't notify us again. You can avoid creating draft if it's not really necessary for testing or specific review of someone for something, because we don't read draft anymore by default

src/components/Imports/Providers/nextcloud/accountService.js
```diff
@@ -0,0 +1,31 @@
...
+      password,
+      url
+    },
+    identifier: 'login',
+    name: login,
+    state: null
+  }
+
+  const res = await client.save(accountData)
+  // cozy-client usually returns the saved doc directly, but keep a safe fallback
```
Crash--: We should be confident with cozy-client. No need to have "safe fallback". If it worked => you'll have res.data, if not you'll have an exception

src/components/Imports/Providers/nextcloud/accountService.js
```diff
@@ -0,0 +1,31 @@
...
+  return res && res.data ? res.data : res
+}
+
+export async function listNextcloudAccounts(client) {
```
Crash--: Not used? 

Also take a look at our naming convention https://github.com/cozy/cozy-guidelines?tab=readme-ov-file#naming-of-functions 

We should avoid the call to fetchJSON, client.query is our friend.

src/components/Imports/Providers/nextcloud/provider.js
```diff
@@ -0,0 +1,449 @@
...
+
+      const txt = await res.text().catch(() => '')
+      const err = new Error(txt || message)
+      err.status = res.status
+      err.path = clean
+      throw err
+    }
+  },
+
+  async importPathRecursive(
```
Crash--: i don't see where this method is called?

src/components/Imports/Providers/nextcloud/provider.js
```diff
@@ -0,0 +1,449 @@
...
+    if (res.status === 409) {
+      const existing = await this.findChildDirByName(client, parentId, name)
+      if (existing) return existing
+    }
+
+    const txt = await res.text().catch(() => '')
+    throw new Error(txt || `Create directory failed: HTTP ${res.status}`)
+  },
+
+  async ensureChildDir(client, parentId, name) {
```
Crash--: don't we already have `ensureDirectoryExists` or something like that in CozyClient?

src/components/Imports/Providers/nextcloud/provider.js
```diff
@@ -0,0 +1,449 @@
...
+    const safeLogin = String(login || 'unknown').replace(/[/\\]/g, '_')
+    const finalDir = await this.ensureChildDir(
+      client,
+      providerDir._id || providerDir.id,
+      safeLogin
+    )
+    return finalDir._id || finalDir.id
+  },
+
+  async downstreamFile(
```
Crash--: why not using https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/NextcloudFilesCollection.js#L138 ?

src/components/Imports/Run.jsx
```diff
@@ -0,0 +1,715 @@
...
+              </Stack>
+            )}
+          </Stack>
+        )}
+      </Stack>
+    </Page>
+  )
+}
+
+export default Run
```
Crash--: component too big, too much responsability, too much state. Please split it.

src/reducers/import.js
```diff
@@ -9,6 +15,15 @@ import {
...
 } from '@/actions/import'
 
+const enabled = (state = false, action) => {
+  switch (action.type) {
+    case SET_IMPORTS_ENABLED:
+      return !!action.enabled
+    default:
+      return state
+  }
+}
```
Crash--: redux is an old school way to handle state in an app. We should avoid using it, specially for small data.

Crash-- (commented):
I know the PR is not finished, but at the end, you'll create a PR with 3k lines of code, you should not do that. 

We should add code, behind flag, and merge it. Regularly. Because here, it'll be a nightmare to read and if we ask you to change things, it's so easier to change little PR than big ones.

Crash-- (commented):
Better, but still need a lot of work 💪

KillianCourvoisier:
Ok, all i18n translation and usage are done too, I hope everything is fine now. Lemme know

src/components/Imports/Providers/nextcloud/pathUtils.js
```diff
@@ -0,0 +1,23 @@
...
+
+export function extractName(path) {
+  const clean = joinRemotePath(path || '/')
+  const parts = clean.split('/').filter(Boolean)
+  return parts.pop() || '/'
+}
+
+export function build404Reason(path) {
+  const name = extractName(path)
+  const hasWeirdChars = /[^\u0020-\u007E]/.test(name) || /[*'^()]/.test(name)
```
Crash--: only `const ForbiddenFilenameChars = "/\x00\n\r"` are not supported by stack?

Crash-- (commented):
Still super complicated to understand. 

If I have to maintain that by myself, I'm not very confident.

PR #816, opened 2026-04-15
feat: Add warning chip for organization if domain are not configured

JF-Cozy:


src/components/Profile/ProfileView.jsx
```diff
@@ -62,9 +69,16 @@ const ProfileView = ({
...
+            {isEmailReadOnly ? (
+              <EmailReadOnlySection isMailAppInstalled={isMailAppInstalled} />
+            ) : (
+              <EmailSection />
+            )}
             {isMatrixEnabled && (
-              <MatrixIdSection email={instanceResult.data.email} />
+              <MatrixIdSection
+                email={instanceResult.data.email}
+                isChatAppInstalled={isChatAppInstalled}
```
rezk2ll: The render only waits on `instanceResult`, not on `appsResult`. On a cold load, the instance query may resolve first; while that happens, `appsResult.data` is still `null`, so `apps` falls back to `[]`, both booleans are `false`, and `showTag` becomes `true` even for users with the app installed

cleaner option: make `isMailAppInstalled` / `isChatAppInstalled` tri-state (`true` / `false` / `undefined`) and only render the chip when the value is explicitly `false` and the chip disappears a moment later once the apps query resolves.

src/components/Profile/ProfileView.jsx
```diff
@@ -43,6 +43,13 @@ const ProfileView = ({
     instanceQuery.options
   )
 
+  const appsQuery = buildAppsQuery()
+  const appsResult = useQuery(appsQuery.definition, appsQuery.options)
+  const apps = appsResult.data || []
+
+  const isChatAppInstalled = apps.some(app => app.slug.includes('chat'))
```
rezk2ll: this will give an incorrect result if we have another app that contains the word chat or mail.

for example, videochat or chatbot, or mail, tmail, etc..

consider: `app.slug === 'chat'`

src/components/Profile/MatrixIdSection.jsx
```diff
@@ -15,10 +16,14 @@ export const generateMatrixId = email => {
   return `@${username}:${domain}`
 }
 
-const MatrixIdSection = ({ email }) => {
+const MatrixIdSection = ({ email, isChatAppInstalled }) => {
   const { t } = useI18n()
+  const {
+    instance: { data: instanceData }
+  } = useInstanceInfo()
```
rezk2ll, line 23: There is no *functional* bug here, but it is harder to reason about the `useInstanceInfo()` addition in `MatrixIdSection`.

We have 3 different paths to fetch the user info in the profile view:
- `ProfileView` itself:  via `buildSettingsInstanceQuery()` + `useQuery`
- `EmailReadOnlySection` via its own `buildSettingsInstanceQuery()` + `useQuery`
- `MatrixIdSection`: via `useInstanceInfo()`, which also pulls in `context` and `disk-usage` as side effects

Since the parent already has `org_id` on hand, the simplest fix is to pass a single `showTag` (or `isInOrg`) prop down from `ProfileView` the same way `email` and `isChatAppInstalled` are passed today.

src/components/Profile/ProfileView.jsx
JF-Cozy: good catch, changed

src/components/Profile/ProfileView.jsx
JF-Cozy: oh yes you're right, initially the condition was `showTag = !!instance.org_id && isMailAppInstalled` (by mistake) so it worked like this... but not anymore since we have now the correct condition `!isMailAppInstalled` ;)

I changed the approach to use more conventional way by testing the fetch status of each queries. I also extract queries to have a dumber component

src/components/Profile/MatrixIdSection.jsx
JF-Cozy: yes we should harmonize the way to query things, maybe in an other PR...
for the second point, yes we have both method : autonomous comp and comp with props. We should harmonize as well 🤔

rezk2ll (approved):

PR #820, opened 2026-04-20
feat: Add Nextcloud migration progress flow

doubleface:
- Add `useMigration` hook to poll migration progress from the stack and expose state (progress %, elapsed time, cancel logic)
- Add `NextcloudProgressDialog` component: full-screen dialog with Lottie animations (in-progress + done states), linear progress bar, estimated remaining time, and cancel/open-Drive actions
- Wire the progress dialog into the existing Nextcloud migration flow, replacing the static connect dialog after launch

[Capture vidéo du 2026-04-20 16-07-15.webm](https://github.com/user-attachments/assets/e3d01654-8069-44cb-986e-5bded38381c3)


src/components/Migration/useMigration.js
```diff
@@ -0,0 +1,143 @@
...
+    const fetchDriveUrl = async () => {
+      const baseLink = {
+        cozyUrl: client.getStackClient().uri,
+        slug: 'drive',
+        subDomainType
+      }
+      try {
+        const { data: folder } = await client
+          .collection('io.cozy.files')
+          .statByPath('/Nextcloud imported files')
```
zatteo, line 31: Are we sure about this hardcoded in English file path?

zatteo:
So we can only import one Nextcloud account?

src/components/Migration/useMigration.js
```diff
@@ -0,0 +1,143 @@
...
+  const start = useCallback(
+    async (credentials = {}) => {
+      setError(null)
+      setCancelSuccess(false)
+      setIsLoading(true)
+      try {
+        const result = await client
+          .getStackClient()
+          .fetchJSON('POST', '/remote/nextcloud/migration', credentials)
+        setStartedAt(result.data.started_at || new Date().toISOString())
```
zatteo, line 76: Is it really possible that `result.data.started_at` is null?

src/components/Migration/useMigration.js
```diff
@@ -0,0 +1,143 @@
...
+    if (!migrationId) return
+    setIsCanceling(true)
+    try {
+      await client
+        .getStackClient()
+        .fetchJSON('POST', `/remote/nextcloud/migration/${migrationId}/cancel`)
+      setCancelSuccess(true)
+    } catch (e) {
+      if (e.status === 503) setError('cancel_unavailable')
+      // 409 means already terminal: not an error from the user's perspective
```
zatteo, line 98: So it means that if there is another error than 503 it get silently ignored?

src/components/Migration/NextcloudProgressDialog.jsx
```diff
@@ -0,0 +1,212 @@
...
+
+  const isDone =
+    progress &&
+    progress.files_imported >= progress.files_total &&
+    progress.files_total > 0
+
+  const percent =
+    progress && progress.files_total > 0
+      ? Math.round((progress.files_imported / progress.files_total) * 100)
+      : 0
```
zatteo, line 58: nit: To put in another helper like `computeRemainingSeconds` ?

src/components/Migration/Migration.jsx
```diff
@@ -94,7 +94,12 @@ const Migration = () => {
       </Typography>
       <Typography variant="body2" color="textSecondary">
         {t('MigrationView.help.description')}{' '}
-        <a href="#/support" className="u-link">
+        <a
+          href="https://twake.app/en/support/"
```
zatteo, line 98: I think we have a dynamic support link we should use instead of an hardcoded one

src/components/Migration/NextcloudConnectDialog.jsx
```diff
@@ -13,37 +13,55 @@ import TextField from 'cozy-ui/transpiled/react/TextField'
...
 
-  const handleSubmit = async () => {
-    setError(null)
-    setIsLoading(true)
+  const handleSubmit = () =>
+    start({
+      nextcloud_url: url,
+      nextcloud_login: username,
+      nextcloud_app_password: password,
+      target_dir: 'Nextcloud imported files'
```
zatteo, line 42: To put in a const 'Nextcloud imported files' used by the statByFile seen previously?

zatteo:
Centralizing everything in a hook `useMigration` looks like a good idea!

doubleface:
> So we can only import one Nextcloud account?

@zatteo Yes at the moment. It is just a first step

src/components/Migration/useMigration.js
doubleface: I don't think so but it's just defensive code

src/components/Migration/useMigration.js
doubleface: Yes, we should have a more generic error in unexpected cases. Fixed

src/components/Migration/NextcloudProgressDialog.jsx
doubleface: yes, extracted it to useMigration hook also with isMigrationDone

src/components/Migration/Migration.jsx
doubleface: fixed. Now using the help link from instance context

src/components/Migration/useMigration.js
doubleface: Right, let's translate this folder name

src/components/Migration/NextcloudConnectDialog.jsx
doubleface: Finally localized the destination folder

src/components/Migration/NextcloudProgressDialog.jsx
```diff
@@ -0,0 +1,211 @@
...
+                  t
+                )
+                return formattedTime ? (
+                  <Typography variant="body2" color="textSecondary">
+                    {t('MigrationView.nextcloud.progress.remaining', {
+                      time: formattedTime
+                    })}
+                  </Typography>
+                ) : null
+              })()}
```
zatteo: This could be a component?

src/components/Migration/NextcloudConnectDialog.jsx
doubleface: And re finally, we don't localze the destination folder to avoid having multiple dstination folder on change of localization

src/components/Migration/NextcloudProgressDialog.jsx
doubleface: Yes done. Now it's MigrationProgressInfo component

zatteo (approved):

PR #824, opened 2026-04-28
feat: Add nextcloud migration progress banner

doubleface:
- Add a global `MigrationProgressBanner` that follows the active Nextcloud migration in realtime.
- Display a "Migrated" chip and a "Clean Nextcloud" action on the migration page once a migration is `completed`
- For testing purpose, behind `drive.reset-for-migration.enabled` flag "Clean Nextcloud" action
  cleans migration documents and migration destination folder to be able to restart a migration.
- Fix a double slash in the imported files target path
- Unit tests for MigrationProgressBanner component

[Capture vidéo du 2026-04-28 12-01-15.webm](https://github.com/user-attachments/assets/d3d52b3f-18cf-4c8f-8dbf-7283502b166d)

https://www.notion.so/linagora/Drive-Board-20062718bad180d687d1f517b2ed7dda?p=33b62718bad180a982f9f7844825980f&pm=s


src/components/Migration/Migration.jsx
```diff
@@ -19,6 +21,18 @@ import NextcloudMigrationDialog from '@/components/Migration/NextcloudMigrationD
 import Page from '@/components/Page'
 import PageTitle from '@/components/PageTitle'
 
+const NEXTCLOUD_MIGRATIONS_DOCTYPE = 'io.cozy.nextcloud.migrations'
```
JF-Cozy: we should put this in doctypes file

src/components/Migration/Migration.jsx
```diff
@@ -19,6 +21,18 @@ import NextcloudMigrationDialog from '@/components/Migration/NextcloudMigrationD
...
+
+const buildCompletedNextcloudMigrationsQuery = () => ({
```
JF-Cozy: we should put this in queries file

src/components/Migration/Migration.jsx
```diff
@@ -29,6 +43,18 @@ const Migration = () => {
 
   const { context } = useInstanceInfo()
   const helpLink = context?.data?.help_link || 'https://twake.app/en/support/'
+  const completedMigrationsQuery = useMemo(
```
JF-Cozy: useMemo not useful here, you can remove it

src/components/Migration/Migration.jsx
```diff
@@ -29,6 +43,18 @@ const Migration = () => {
...
+    () => buildCompletedNextcloudMigrationsQuery(),
+    []
+  )
+  const { data: completedMigrations, ...completedMigrationsQueryState } =
+    useQuery(
+      completedMigrationsQuery.definition,
+      completedMigrationsQuery.options
+    )
+  const isNextcloudMigrated =
+    !isQueryLoading(completedMigrationsQueryState) &&
```
JF-Cozy: I think we rely mostly on hasQueryBeenLoaded 🤔

src/components/Migration/Migration.jsx
```diff
@@ -69,17 +95,44 @@ const Migration = () => {
                   <ProviderLogo icon={provider.icon} alt={provider.name} />
                 </ListItemIcon>
                 <ListItemText
-                  primary={provider.name}
+                  primary={
```
JF-Cozy: maybe it's time to extract this in ProviderItem or something like that to keep cleaner and lighter return content, and maybe you can create a ProviderItemNexctloudMigrated or somehting like that too.

By the way why do we bother with providers array? There is only one single value in it...

src/components/Migration/MigrationProgressBanner.jsx
```diff
@@ -0,0 +1,173 @@
...
+        setMigrationDoc(null)
+        setShowSnackbar(true)
+      } else if (doc.status === 'cancelled' || doc.status === 'error') {
+        setMigrationDoc(null)
+      } else {
+        setMigrationDoc(doc)
+      }
+    }
+
+    client.plugins.realtime.subscribe(
```
JF-Cozy, line 43: we can use Realtime component in upper level (pretty sure we already use it somewhere) to use realtime on this doctype.

src/components/Migration/MigrationProgressBanner.jsx
```diff
@@ -0,0 +1,173 @@
...
+  const [migrationDoc, setMigrationDoc] = useState(null)
+  const [showSnackbar, setShowSnackbar] = useState(false)
+  const [isCanceling, setIsCanceling] = useState(false)
+  const [completedFilesTotal, setCompletedFilesTotal] = useState(0)
+
+  useEffect(() => {
+    let cancelled = false
+    const fetchRunning = async () => {
+      const { data } = await client.query(
+        Q('io.cozy.nextcloud.migrations')
```
JF-Cozy: I think we can use useQuery in a wrapper and pass the data to MigrationProgressBanner

src/components/Migration/MigrationProgressBanner.jsx
JF-Cozy, line 1: we have ProgressionBanner in cozy-ui, why not using it?

src/components/Migration/MigrationProgressBanner.jsx
```diff
@@ -0,0 +1,173 @@
...
+                  />
+                </div>
+              </div>
+              <LinearProgress variant="determinate" value={percent} />
+            </div>
+          </div>
+        </div>
+      )}
+      {showSnackbar && (
+        <Snackbar open onClose={() => setShowSnackbar(false)}>
```
JF-Cozy: we can use useAlert from cozy-ui to avoid creating a specific snackbar

src/components/Migration/MigrationProgressBanner.jsx
```diff
@@ -0,0 +1,173 @@
...
+  const progress = migrationDoc?.progress
+  const percent =
+    progress?.bytes_total > 0
+      ? Math.round((progress.bytes_imported / progress.bytes_total) * 100)
+      : 0
+
+  return (
+    <>
+      {migrationDoc && progress && (
+        <div className="u-p-half">
```
JF-Cozy: I think we can reduce the usage of div and specific style (even with u-classes) by using Alert or Paper component. There is an example in Banner/Readme.md in cozy-ui of Banner creation using Alert comp

src/lib/flags.js
```diff
@@ -12,5 +12,6 @@ export const initFlags = () => {
     flag('settings.specify-blocking-subscription-vendor-dev')
     flag('settings.2fa.enabled')
     flag('settings.delete-user-from-signup-admin-api')
+    flag('drive.reset-for-migration.enabled')
```
JF-Cozy: if it's a cross app flag, maybe we should rename it, or use a settings specific flags here

JF-Cozy (commented):
please squash and refactors commit before sending them to review, it will be easier, more relevant and less time consuming

src/components/Migration/MigrationProgressBanner.jsx
doubleface: I did not know that one. Let's use it

src/components/Migration/Migration.jsx
doubleface: There are multiple providers planned. But maybe we don't need to anticipate this yet.

src/lib/flags.js
doubleface: I will remove this flag once the whole migration is implemented with the real clean actions. This this flag will disappear. I think we don't need to worry about this naming too much.

JF-Cozy: It was, and still is, difficult to track which flag do what and where. We have a doc for that, but naming is, in my opinion, useful to help us for that...

src/components/Migration/MigrationProgressBanner.jsx
```diff
@@ -0,0 +1,132 @@
...
+            busy={isCanceling}
+            data-testid="migration-progress-banner-cancel"
+          />
+        </>
+      }
+    />
+  )
+}
+
+const DumbMigrationProgressBannerMemo = memo(DumbMigrationProgressBanner)
```
JF-Cozy: why memoing here?

src/components/Migration/MigrationProgressBanner.jsx
```diff
@@ -0,0 +1,132 @@
...
+      client.plugins.realtime.unsubscribe(
+        'updated',
+        NEXTCLOUD_MIGRATIONS_DOCTYPE,
+        migrationId,
+        handleUpdate
+      )
+    }
+  }, [client, migrationId, showAlert, t])
+
+  const handleCancel = useCallback(async () => {
```
JF-Cozy, line 59: we have useFetchJSON, maybe useful to avoid doing everything manually?

src/components/Migration/Migration.jsx
```diff
@@ -16,28 +23,86 @@ import Typography from 'cozy-ui/transpiled/react/Typography'
...
 import PageTitle from '@/components/PageTitle'
+import { NEXTCLOUD_MIGRATIONS_DOCTYPE } from '@/doctypes'
+import logger from '@/lib/logger'
+import { buildCompletedNextcloudMigrationsQuery } from '@/lib/queries'
+
+const RESET_NEXTCLOUD_MIGRATION_FLAG = 'drive.reset-for-migration.enabled'
+
+const completedMigrationsQuery = buildCompletedNextcloudMigrationsQuery()
+
+// Temporary test helper: remove with RESET_NEXTCLOUD_MIGRATION_FLAG once the reset feature is fully implemented.
```
JF-Cozy: nit: we should isolate in a file

src/components/Migration/Migration.jsx
```diff
@@ -16,28 +23,86 @@ import Typography from 'cozy-ui/transpiled/react/Typography'
...
+    useState(false)
 
   const { context } = useInstanceInfo()
   const helpLink = context?.data?.help_link || 'https://twake.app/en/support/'
+  const { data: completedMigrations, ...completedMigrationsQueryState } =
+    useQuery(
+      completedMigrationsQuery.definition,
+      completedMigrationsQuery.options
+    )
+  useEffect(() => {
```
JF-Cozy: why using useEffect and not extracting the query in a wrapper?

src/lib/flags.js
doubleface: Let's go with settings specific flag then

src/components/Migration/MigrationProgressBanner.jsx
doubleface: I did not know about useFetchJSON, but looks like it is not adapted in this case since I want to call cancel endboint only when the button is clicked. We cannot call hooks conditionally

src/components/Migration/MigrationProgressBanner.jsx
doubleface: I think it avoided a problem which does not exist anymore -> removed

src/components/Migration/Migration.jsx
```diff
@@ -106,4 +139,57 @@ const Migration = () => {
...
+  const { context } = useInstanceInfo()
+  const helpLink = context?.data?.help_link || 'https://twake.app/en/support/'
+
+  const { data: completedMigrations, ...completedMigrationsQueryState } =
+    useQuery(
+      completedMigrationsQuery.definition,
+      completedMigrationsQuery.options
+    )
+
+  useEffect(() => {
```
JF-Cozy: why this is needed ? 🤔

src/components/Migration/Migration.jsx
```diff
@@ -106,4 +139,57 @@ const Migration = () => {
...
+  }, [client, isCleaningNextcloud])
+
+  return (
+    <DumbMigration
+      helpLink={helpLink}
+      hasCompletedNextcloudMigration={hasCompletedNextcloudMigration}
+      isCleaningNextcloud={isCleaningNextcloud}
+      showNextcloudDialog={showNextcloudDialog}
+      onCleanNextcloud={handleCleanNextcloud}
+      onShowNextcloudDialog={setShowNextcloudDialog}
```
JF-Cozy, line 183: we have multiple props here just for one case, maybe we can create a new component just for those, or use nextcloudProps object wrapper for better storing things...

src/components/Migration/resetNextcloudMigration.js
JF-Cozy, line 1: should be helpers.js file I suppose

JF-Cozy (approved):

src/components/Migration/Migration.jsx
doubleface: no good reason. Removed

PR #825, opened 2026-05-04
feat(migration): Add nextcloud cleanup flow

doubleface:
Let migrated users remove data from their Nextcloud source/

Add confirmation and progress dialogs with localized feedback so users understand the deletion impact and result.

[Capture vidéo du 2026-05-04 09-54-23.webm](https://github.com/user-attachments/assets/66699d90-9d48-4002-b661-4698748fb372)

### DIFF
--- a/src/components/Migration/Migration.jsx
+++ b/src/components/Migration/Migration.jsx
@@ -1,4 +1,4 @@
-import React, { useState, useCallback, useEffect } from 'react'
+import React, { useState, useCallback } from 'react'
 import { useI18n } from 'twake-i18n'
 
 import { useInstanceInfo, useQuery, isQueryLoading } from 'cozy-client'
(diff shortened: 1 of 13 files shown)


src/components/Migration/Migration.jsx
```diff
@@ -140,47 +146,84 @@ const DumbMigration = ({
...
+    if (!cleanError) return
+
+    showAlert({
+      title: t('MigrationView.cleanErrorSnackbar.title'),
+      message: t('MigrationView.cleanErrorSnackbar.description'),
+      severity: 'error',
+      duration: SNACKBAR_AUTO_HIDE_MS
+    })
+    setCleanError(null)
+  }, [cleanError, setCleanError, showAlert, t])
```
zatteo: What do you think of calling `showAlert` after the result of `await clean()` depending of its result to avoid two `useEffect` ?

zatteo:
Video preview is not available.

src/components/Migration/Migration.jsx
zatteo: Or maybe with `onCleanSuccess` and `onCleanError` callbacks in `useMigration`.

doubleface:
> Video preview is not available.

@zatteo Fixed now, sorry

src/components/Migration/Migration.jsx
doubleface: Right, did a refactor which avoids all those useEffect and works well

zatteo: Nice! A lot simplier!

zatteo (approved):

PR #826, opened 2026-05-04
fix: Use correct 48x48 migration icons

doubleface:
image


rezk2ll:
Why we don't have the icon in cozy UI

zatteo (approved):

zatteo:
> Why we don't have the icon in cozy UI

If it's used only here why not?

JF-Cozy:
> Why we don't have the icon in cozy UI

no sens to have those globally I think.

BTW @doubleface not part of the PR but the Listitems doesn't seem to be fullwidth 🤔 don't know if it's intended but generally they are in a modal I think

doubleface:
> > Why we don't have the icon in cozy UI
> 
> no sens to have those globally I think.
> 
> BTW @doubleface not part of the PR but the Listitems doesn't seem to be fullwidth 🤔 don't know if it's intended but generally they are in a modal I think

It's in a modal here and it looks like it is the dialog content which makes them this size

PR #829, opened 2026-05-07
feat(migration): Run nextcloud connector

doubleface:
Add a dedicated hook to create or reuse Nextcloud accounts and
triggers, launch jobs, and expose progress and result state for the
migration UI.

Run nextcloud connector directly from settings apps with migration progression UI
Did not use cozy-harvest-lib to have a better UX integration with existing migration
screens

[Capture vidéo du 2026-05-04 09-54-23.webm](https://github.com/user-attachments/assets/82b713a3-a86c-4414-a80e-071ec12593fe)


zatteo:
I don't understand this PR. I thought there was no link between the Nextcloud migration process and Nextcloud konnector? I misunderstood?

doubleface:
> I don't understand this PR. I thought there was no link between the Nextcloud migration process and Nextcloud konnector? I misunderstood?

@zatteo There are actually 2 path of "migration" :
- the "transfer" which is what you have eared of, which is done by a specific migration server and does not use the existing nextcloud connector
- the "connection", which uses the nextcloud connector to create an external drive

And to have a better UX, we have intergrated the launch of the connector directly from the settings application (with progress animation and all) without redirecting to home application

doubleface:
Finally, we won't need this feature

PR #831, opened 2026-05-11
feat: Optimize eslint execution time with cache and concurrency

doubleface:
See https://github.com/linagora/twake-drive/pull/3838 for details

## Summary by CodeRabbit

* **Chores**
  * Updated linting script configuration to run with concurrent processing and content-based caching.
  * Refined repository ignore rules to exclude additional editor and build-related files.

[Review Change Stack](https://app.coderabbit.ai/change-stack/linagora/cozy-settings/pull/831)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,10 +20,13 @@ desktop.ini
 .floo
 .flooignore
 .vscode
+.opencode
 
 # Default
 !.gitkeep
 
 # SWC
 .swc
-.opencode
+
+# eslint cache
+.eslintcache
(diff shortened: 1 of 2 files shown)


rezk2ll (approved):

JF-Cozy:
@doubleface why did we need to do that? What's the issue we solved here? Is it worth to use it globally?

doubleface:
> @doubleface why did we need to do that? What's the issue we solved here? Is it worth to use it globally?

There are more justifications on the linked https://github.com/linagora/twake-drive/pull/3838 PR:

- with cache, the lint is a lot faster: something like from 20s to 3s on drive and if agents use yarn lint it makes them a lot faster
- with concurrency, first run is a little bit faster (something from 20s to 15s), which is objectively better. It also makes ci a little bit faster.
- maybe there is a better way to share it amongst project but I did not find it yet.

PR #841, opened 2026-06-04
fix(settings): mark password as managed by an external provider

rezk2ll:
## Problem

- On instances that delegate credentials to an external identity provider, the password section had no good state: with a local password it offered the inline Cozy form (which cannot change anything there), and with no change URL it fell through to nothing useful.
- The distinguishing condition cannot be read from `hasPassword` or `settings.password.readonly`: the cloudery provisions a default password (so one always exists) and `readonly` is typically not set for these instances.

## Solution

- Add a `settings.password.external` flag meaning the password is managed by an external identity provider.
- When set, the section shows explanatory text and no button, independent of `hasPassword` and `readonly`.
- The existing paths are untouched: `settings.password.url` still renders the external link button, and the native inline form (with its `readonly` handling) is unchanged.
- Adds the `ProfileView.password.external_label` string in English, French, Russian and Vietnamese.

| Outcome | Condition |
| --- | --- |
| External link button | `settings.password.url` is set |
| "Managed by your identity provider" text, no button | `settings.password.external` is set |
| Inline Cozy form (disabled when read-only) | neither set, user has a password |
| Nothing | neither set, no password |


zatteo (approved):
Make sense

rezk2ll:
i will hold off on this PR until i have the proper wording etc from the PO

Crash--:
@rezk2ll please find a way to merge & deploy this :)

rezk2ll:
> @rezk2ll please find a way to merge & deploy this :)

i got confirmation from the PO now

PR #843, opened 2026-06-10
feat: Migrate to Yarn 4.16

zatteo:
.yarnrc.yml has been automatically generated to be non breaking

## Chores

* Updated package manager settings to enforce Yarn 4.16.0 across the project.
* Adjusted the build workflow to enable Corepack and switched dependency installation to use immutable lockfile validation.
* Updated Yarn configuration to use the `node-modules` linker for installed dependencies.
* Expanded `.gitignore` to better cover Yarn Plug’n’Play artifacts, caches, and related directories.


.yarnrc.yml
```diff
@@ -1,8 +1 @@
-approvedGitRepositories:
-  - "**"
-
-enableScripts: true
-
 nodeLinker: node-modules
-
-npmMinimalAgeGate: 0
```
zatteo: Some thought about this? Should I be explicit and set the same values (maybe the same as default but at least to be explicit) ?

rezk2ll:
@zatteo can this be merged / closed?

.yarnrc.yml
zatteo: Going with implicit here.

